PR #2, opened 2020-06-18
build & configuration of machine

benadida:
admin scripts in `tty2` are not fully working yet, some path issues to resolve with new setup, but everything else is in good shape.


config/xinitrc
```diff
@@ -0,0 +1,24 @@
...
+
+unclutter -idle 0.01 -root &
+
+if [ "${VX_MACHINE_TYPE}" = bmd ]; then
+    URL=http://localhost:3000/speech-loader.html
+else
+    URL=http://localhost:3000/
+fi
+
+exec sh <<SHELL
```
eventualbuddha, line 15: This works? I’d have expected `exec sh <<SHELL` for it to be terminated by `SHELL` later.

run-election-manager.sh
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
 
+# go to directory where this file is located
+cd "$(dirname "$0")"
```
eventualbuddha, line 4: This sometimes gives the wrong results, like when symlinks are involved. I’m guessing it’s fine for us?

setup-machine.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
```
eventualbuddha, line 2: `set -euo pipefail`? Safety First™

eventualbuddha: Also on the other scripts like `setup-lockdown.sh`.

config/xinitrc
benadida: I think something got mangled here. I'll check :)

run-election-manager.sh
benadida: Yeah I think it's okay, but if you have a better suggestion, I'm not a shell master and welcome it.

config/xinitrc
benadida: no, actually, not mangled, yes this works. I'm not sure I understand what you think it might do?

eventualbuddha: Oh. Weird. This was an issue with the Github iOS app. It only showed `exec sh <`.

run-election-manager.sh
eventualbuddha: Here's a SO post on the subject: https://stackoverflow.com/a/246128/549363. The more correct version is… verbose. Let's just go with this for now.

PR #38, opened 2020-10-11
fix the way sound is set up on BMD

benadida:


run-kiosk-browser-forever-and-log.sh
```diff
@@ -17,7 +17,8 @@ fi
     
 
 if [ "${VX_MACHINE_TYPE}" = "bmd" ]; then
-    amixer set Master 75%
+    amixer -D pulse set Master 80% || true
```
eventualbuddha, line 20: Where does the `pulse` device name come from?

eventualbuddha (approved):

run-kiosk-browser-forever-and-log.sh
benadida: as best as I understand, it is the standard pulseaudio device on Linux. For some reason on the hardware we're using, maybe because we don't have gnome installed to clean things up, not specifying a device causes the command to fail. And I don't have an easy way to look up the device names either that I've seen consistently work.

PR #39, opened 2020-10-11
bumps for scanner

benadida:


benadida:
@eventualbuddha just bumps, approve post-facto pretty please :)

PR #48, opened 2020-11-17
Increase cursor size

carolinemodic:
Increases the size of the cursor to 32px from the default of 16px. 

Before: 
cursorsizedefault

After: 
cursorsize32

For context in case we want to go even bigger this is what 64px looks like: 
curssorsize64

Tested by running the setup-machine script on a fresh ubuntu virtual machine and setting up as election manager and verifying the cursor size changed and persisted when opening file pickers, etc. I noticed (even without these changes) the cursor also exists when set up as bmd, I assume the touch screen laptops somehow remove the cursor outside of what I can test on my VM but I wasn't sure exactly how that works.

### DIFF
--- a/config/Xresources
+++ b/config/Xresources
@@ -1,0 +1,1 @@
+Xcursor.size: 32
--- a/config/xinitrc
+++ b/config/xinitrc
@@ -1,6 +1,7 @@
 xset -dpms
 xset s off
 xset s noblank
+xrdb ~/.Xresources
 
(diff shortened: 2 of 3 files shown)


benadida:
`unclutter` in `run-kiosk-forever-and-log.sh` is how we remove the cursor. I wonder if this change conflicts with it? At the same time, in the VM, I'm pretty sure it's hard to tell cause of course the mouse is still visible, so I'm good to merge this.

benadida (approved):

eventualbuddha (approved):

PR #50, opened 2020-11-18
Tweak the GtkFileChooserDialog settings

carolinemodic:
Makes a couple of small tweaks to the file chooser dialog. 

1. For all gtk dialogs (file opener, file saver, error message dialogs) increases the font size. 
2. Removes the "recent" directory that shows up in the left hand side when you are trying to load a file. This is the only thing in the left hand drive selection area you can configure so I figured we should remove it.  
3. Removes the openbox blue title bar from the file chooser dialog (but not other gtk dialogs like error messages). I thought this looked weird so I removed it, but open to adding it back. Part of why it looks weird right now is just how we're opening the dialogs in kiosk-browser is not sending in a title that we could be populating which might make it look a little better. FWIW we can also adjust the color/font size of this. 
4. Sets the file choose dialog to be larger in the screen, since we increase the font size it makes sense to make the whole dialog bigger. 

Screenshots are from have parallels fullscreen on a monitor that is 1920x1080 pixels which is the same as our laptop resolution so I think it should be representative. 
# Before Screenshots: 
## Loading a file
beforeopenfile
## Saving a file
beforesavefile
## Error Dialog
errordialo

# After Screenshots:
## Loading a file
loadingafter
## Saving a file
savingafter
## Error Dialog
errordialogafter


eventualbuddha (approved):

beausmith (approved):
This is great!

benadida (approved):
this is a great immediate improvement.

PR #54, opened 2021-01-21
fix: standardize run scripts

eventualbuddha:
- `VX_CONFIG_ROOT` is not the `config` directory; it's where the config files are stored i.e. `/vx-config`
- add `set -euo pipefail` "strict mode" directive to scripts
- make `run-*.sh` scripts executable
- set `NODE_ENV=production`

### DIFF
--- a/run-bas.sh
+++ b/run-bas.sh
@@ -1,12 +1,13 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 # go to directory where this file is located
 cd "$(dirname "$0")"
 
-: "${VX_CONFIG_ROOT:="./config"}"
-
 # configuration information
-source ${VX_CONFIG_ROOT}/read-vx-machine-config.sh
+source config/read-vx-machine-config.sh
 
 export PIPENV_VENV_IN_PROJECT=1
+export NODE_ENV=production
 (trap 'kill 0' SIGINT SIGHUP; make -C vxsuite/apps/module-smartcards run & make -C vxsuite/apps/bas run)
--- a/run-bmd.sh
+++ b/run-bmd.sh
@@ -1,12 +1,13 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 # go to directory where this file is located
 cd "$(dirname "$0")"
 
-: "${VX_CONFIG_ROOT:="./config"}"
-
 # configuration information
-source ${VX_CONFIG_ROOT}/read-vx-machine-config.sh
+source config/read-vx-machine-config.sh
 
 export PIPENV_VENV_IN_PROJECT=1
+export NODE_ENV=production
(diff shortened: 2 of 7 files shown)


run-bas.sh
```diff
@@ -1,12 +1,15 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 # go to directory where this file is located
 cd "$(dirname "$0")"
 
-: "${VX_CONFIG_ROOT:="./config"}"
+: "${VX_CONFIG_ROOT:="/vx-config"}"
```
benadida, line 8: might this line no longer be needed?

run-bmd.sh
```diff
@@ -1,12 +1,15 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 # go to directory where this file is located
 cd "$(dirname "$0")"
 
-: "${VX_CONFIG_ROOT:="./config"}"
+: "${VX_CONFIG_ROOT:="/vx-config"}"
```
benadida, line 8: same, maybe no longer necessary?

run-bsd.sh
```diff
@@ -1,12 +1,15 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 # go to directory where this file is located
 cd "$(dirname "$0")"
 
-: "${VX_CONFIG_ROOT:="./config"}"
+: "${VX_CONFIG_ROOT:="/vx-config"}"
```
benadida, line 8: same x 3

run-election-manager.sh
```diff
@@ -5,10 +5,11 @@ set -euo pipefail
 # go to directory where this file is located
 cd "$(dirname "$0")"
 
-: "${VX_CONFIG_ROOT:="./config"}"
+: "${VX_CONFIG_ROOT:="/vx-config"}"
```
benadida, line 8: x 4 ;)

run-kiosk-browser-forever-and-log.sh
```diff
@@ -2,12 +2,14 @@
 
 set -euo pipefail
 
-# load configuration
-source /vx-config/read-vx-machine-config.sh
-
 # go to directory where this file is located
 cd "$(dirname "$0")"
 
+: "${VX_CONFIG_ROOT:="/vx-config"}"
```
benadida, line 8: same x5 ;)

benadida (approved):
possibly useless line to remove, otherwise good.

run-bas.sh
eventualbuddha: I could remove this since it's set in production by the `.service` files. It's probably worth putting in an explicit check for it, though, for testing. What do you think?

benadida: I'm not seeing used at all in this file, though. Are you setting it for the other scripts we're sourcing? Wouldn't those other scripts be better positioned to set a default?

eventualbuddha: You're right. Will remove.

PR #56, opened 2021-02-03
Bumps + Improvements in file structure in production + fixes to config loading

benadida:
As title says :)

### DIFF
--- a/setup-machine.sh
+++ b/setup-machine.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# /vx/ui --> home directory for the vx-ui user (rw with symlink for ro files)
+# /vx/services --> home directory for the vx-services user (rw with symlink for ro files)
+# /vx/admin --> home directory for the vx-admin user (rw with symlink for ro files)
+# /vx/code --> all the executable code (ro)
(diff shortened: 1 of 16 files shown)


carolinemodic (approved):

benadida:
confirmed that `election-manager` and `bmd` work as full installs on a VM with these changes.

setup-machine.sh
```diff
@@ -45,10 +52,15 @@ sudo cp config/20auto-upgrades /etc/apt/apt.conf.d/
 # make sure machine never shuts down on idle, and does shut down on power key (no hibernate or anything.)
 sudo cp config/logind.conf /etc/systemd/
 
+# directory structure
+sudo mkdir /vx
+sudo mkdir /vx/code
+sudo mkdir -p /vx/data/module-scan
```
eventualbuddha, line 58: Will this be writable by `vx-*`? I don't see a `chown` for it.

eventualbuddha (approved):

setup-machine.sh
benadida: in fact no, I just tried ballot-scanner and it failed :) Will fix.

benadida: fixed!

PR #74, opened 2021-05-09
fix: don't remove module-smartcards in EM anymore

benadida:


benadida:
as I need to build an image today, I'm going to merge this very small change and look for post-facto approval @eventualbuddha @carolinemodic hope that's okay!

PR #75, opened 2021-05-12
feat(scan): add `precinct-scanner` app

eventualbuddha:
This pins vxsuite to the latest draft version that includes the `precinct-scanner` prototype.


config/vx-precinct-scanner.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=VotingWorks Precinct Scanner
+
+[Service]
+Type=simple
+User=vx-services
+Environment=VX_CONFIG_ROOT=/vx/config
```
benadida, line 7: should this also have an environment variable for `VX_MACHINE_TYPE` ?

benadida: and then for the `bsd` service, the other value?

benadida (approved):
one question.

config/vx-precinct-scanner.service
eventualbuddha: This happens automatically because of this line in `run-*.sh`:

```sh
source ${CONFIG}/read-vx-machine-config.sh
```

benadida: oh duh

benadida (approved):

benadida:
I've reviewed @eventualbuddha's updates, and I've bumped modules, so merging now.

PR #76, opened 2021-05-20
bump module converter sems

carolinemodic:
Fixes https://zube.io/votingworks/vxsuite/c/3299 and brings in manual data entry changes


benadida (approved):
oh geez! Good catch.

PR #84, opened 2021-08-09
fix: allow alphanumeric and dashes in machine ID

eventualbuddha:
Per [@adghayes](https://votingworks.slack.com/archives/CEL6D3GAD/p1628532274070400?thread_ts=1625587649.004000&cid=CEL6D3GAD):

> @adghayes: when I try to change it in tty2, I get a validation error saying that only digits are allowed
> @eventualbuddha: what should we allow?
> @adghayes: alphanumeric + dashes, I think

### DIFF
--- a/config/admin-functions/choose-vx-machine-id.sh
+++ b/config/admin-functions/choose-vx-machine-id.sh
@@ -6,12 +6,12 @@ set -euo pipefail
 
 while true; do
   read -p "${VX_MACHINE_TYPE} Machine ID (e.g. 0012): " MACHINE_ID
-  if [[ "${MACHINE_ID}" =~ ^[0-9]+$ ]]; then
+  if [[ "${MACHINE_ID}" =~ ^[-0-9A-Z]+$ ]]; then
     mkdir -p "${VX_CONFIG_ROOT}"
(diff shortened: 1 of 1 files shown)


config/admin-functions/choose-vx-machine-id.sh
```diff
@@ -6,12 +6,12 @@ set -euo pipefail
 
 while true; do
   read -p "${VX_MACHINE_TYPE} Machine ID (e.g. 0012): " MACHINE_ID
-  if [[ "${MACHINE_ID}" =~ ^[0-9]+$ ]]; then
+  if [[ "${MACHINE_ID}" =~ ^[-0-9A-Z]+$ ]]; then
```
benadida, line 9: does the first `-` need to be backslashed? I hate regexps.

benadida (approved):

config/admin-functions/choose-vx-machine-id.sh
eventualbuddha: > does the first `-` need to be backslashed?

[No](https://regexr.com/63sb7):

image

> I hate regexps.

This is expected behavior.

PR #87, opened 2021-09-13
TOTP support

benadida:
sudo permissions and install script

### DIFF
--- a/config/sudoers
+++ b/config/sudoers
@@ -19,8 +19,9 @@ Defaults	secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/b
 # User privilege specification
 root	ALL=(ALL:ALL) ALL
 
-# Allow vx-admin to set hostname and time
+# fine-grained sudo permissions for certain users & actions
 vx-admin ALL=(root:ALL) NOPASSWD: /usr/bin/hostnamectl
 vx-admin ALL=(root:ALL) NOPASSWD: /usr/bin/timedatectl
 vx-ui ALL=(root:ALL) NOPASSWD: /usr/bin/timedatectl
 vx-ui ALL=(root:ALL) NOPASSWD: /usr/bin/brightnessctl
+vx-ui ALL=(root:ALL) NOPASSWD: /usr/local/bin/tpm2-totp
--- a/setup-tpm2-totp.sh
+++ b/setup-tpm2-totp.sh
@@ -1,7 +1,7 @@
 set -euo pipefail
 
 # install core dependencies
-sudo apt -y install build-essential autoconf autoconf-archive automake m4 libtool gcc pkg-config libqrencode-dev pandoc doxygen liboath-dev iproute2 plymouth libplymouth-dev libssl-dev libjson-c-dev libcurl4-openssl-dev
+sudo apt -y install build-essential autoconf autoconf-archive automake m4 libtool gcc pkg-config libqrencode-dev plymouth libplymouth-dev libssl-dev libjson-c-dev libcurl4-openssl-dev
 
 (
     cd tpm2-software/tpm2-tss
--- a/setup-tpm2-totp.sh
+++ b/setup-tpm2-totp.sh
@@ -22,6 +22,9 @@ sudo apt -y install build-essential autoconf autoconf-archive automake m4 libtoo
 # reindex shared objects
 sudo ldconfig
 
+# clear out any preexisting TPM-bound TOTP
+sudo tpm2-totp clean
+
 # initialize the TOTP code, which will display the QR code with the secret.
 sudo tpm2-totp --pcrs=0,7 init
 
--- a/setup-tpm2-totp.sh
+++ b/setup-tpm2-totp.sh
@@ -1,32 +1,29 @@
 set -euo pipefail
 
+# install core dependencies
 sudo apt -y install build-essential autoconf autoconf-archive automake m4 libtool gcc pkg-config libqrencode-dev pandoc doxygen liboath-dev iproute2 plymouth libplymouth-dev libssl-dev libjson-c-dev libcurl4-openssl-dev
 
-mkdir ./tpm2-totp
-cd tpm2-totp
-
-git clone https://github.com/tpm2-software/tpm2-tss
 (
-cd tpm2-tss
-./bootstrap
-./configure
-make 
-sudo make install
+    cd tpm2-software/tpm2-tss
+    ./bootstrap
+    ./configure
+    make 
+    sudo make install
 )
 
-git clone https://github.com/tpm2-software/tpm2-totp
-cd tpm2-totp
-./bootstrap
-./configure
-make
-sudo make install
+(
+    cd tpm2-software/tpm2-totp
+    ./bootstrap
+    ./configure
+    make
+    sudo make install
+)
 
+# reindex shared objects
 sudo ldconfig
 
+# initialize the TOTP code, which will display the QR code with the secret.
 sudo tpm2-totp --pcrs=0,7 init
 
-cd ../
-
-rm -rf tpm2-tss tpm2-totp
 
 
--- a/setup-tpm2-totp.sh
+++ b/setup-tpm2-totp.sh
@@ -1,3 +1,4 @@
+set -euo pipefail
 
 sudo apt -y install build-essential autoconf autoconf-archive automake m4 libtool gcc pkg-config libqrencode-dev pandoc doxygen liboath-dev iproute2 plymouth libplymouth-dev libssl-dev libjson-c-dev libcurl4-openssl-dev
 
@@ -5,12 +6,13 @@ mkdir ./tpm2-totp
 cd tpm2-totp
 
 git clone https://github.com/tpm2-software/tpm2-tss
+(
 cd tpm2-tss
 ./bootstrap
 ./configure
 make 
 sudo make install
-cd ..
+)
 
 git clone https://github.com/tpm2-software/tpm2-totp
 cd tpm2-totp
--- a/setup-tpm2-totp.sh
+++ b/setup-tpm2-totp.sh
@@ -1,0 +1,30 @@
+
+sudo apt -y install build-essential autoconf autoconf-archive automake m4 libtool gcc pkg-config libqrencode-dev pandoc doxygen liboath-dev iproute2 plymouth libplymouth-dev libssl-dev libjson-c-dev libcurl4-openssl-dev
+
+mkdir ./tpm2-totp
+cd tpm2-totp
+
+git clone https://github.com/tpm2-software/tpm2-tss
+cd tpm2-tss
+./bootstrap
+./configure
+make 
+sudo make install
+cd ..
+
+git clone https://github.com/tpm2-software/tpm2-totp
+cd tpm2-totp
+./bootstrap
+./configure
+make
+sudo make install
+
+sudo ldconfig
+
+sudo tpm2-totp --pcrs=0,7 init
+
+cd ../
+
+rm -rf tpm2-tss tpm2-totp
+
+
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,9 @@
 [submodule "vxsuite"]
 	path = vxsuite
 	url = https://github.com/votingworks/vxsuite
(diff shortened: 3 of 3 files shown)


setup-tpm2-totp.sh
```diff
@@ -0,0 +1,30 @@
+
+sudo apt -y install build-essential autoconf autoconf-archive automake m4 libtool gcc pkg-config libqrencode-dev pandoc doxygen liboath-dev iproute2 plymouth libplymouth-dev libssl-dev libjson-c-dev libcurl4-openssl-dev
```
umbernhard, line 2: Fwiw pandoc, doxygen, liboath-dev, and I think iproute2 are dev-only dependencies, which I'm pretty sure we don't need.

setup-tpm2-totp.sh
```diff
@@ -0,0 +1,30 @@
...
+
+mkdir ./tpm2-totp
```
umbernhard, line 4: Does it make sense to have this as the top-level directory? I don't think it matters that much, but it's a little odd to have `tpm2-totp/tpm2-tss` and `tpm2-totp/tpm2-totp`. would it be better to make something like `tpm2-software` top directory?

setup-tpm2-totp.sh
```diff
@@ -0,0 +1,30 @@
...
+git clone https://github.com/tpm2-software/tpm2-totp
+cd tpm2-totp
+./bootstrap
+./configure
+make
+sudo make install
+
+sudo ldconfig
+
+sudo tpm2-totp --pcrs=0,7 init
```
umbernhard, line 24: What's the plan for enrollment here? Maybe a wait for input or something to make sure the QR code gets scanned?

umbernhard (approved):

config/sudoers
```diff
@@ -19,8 +19,9 @@ Defaults	secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/b
 # User privilege specification
 root	ALL=(ALL:ALL) ALL
 
-# Allow vx-admin to set hostname and time
+# fine-grained sudo permissions for certain users & actions
 vx-admin ALL=(root:ALL) NOPASSWD: /usr/bin/hostnamectl
 vx-admin ALL=(root:ALL) NOPASSWD: /usr/bin/timedatectl
 vx-ui ALL=(root:ALL) NOPASSWD: /usr/bin/timedatectl
 vx-ui ALL=(root:ALL) NOPASSWD: /usr/bin/brightnessctl
+vx-ui ALL=(root:ALL) NOPASSWD: /usr/local/bin/tpm2-totp
```
eventualbuddha, line 27: Are there any sub-commands with this that would be bad to allow?

setup-tpm2-totp.sh
```diff
@@ -0,0 +1,30 @@
+
+sudo apt -y install build-essential autoconf autoconf-archive automake m4 libtool gcc pkg-config libqrencode-dev pandoc doxygen liboath-dev iproute2 plymouth libplymouth-dev libssl-dev libjson-c-dev libcurl4-openssl-dev
+
+mkdir ./tpm2-totp
+cd tpm2-totp
+
+git clone https://github.com/tpm2-software/tpm2-tss
```
eventualbuddha, line 7: Not submodules? Doing it this way means we have to pin a version using `git checkout` or similar, which is doable but this PR doesn't do it.

Alternatively, since you are just `rm -rf`ing at the end, use a release download instead e.g. https://github.com/tpm2-software/tpm2-tss/archive/refs/tags/3.1.0.tar.gz?

setup-tpm2-totp.sh
```diff
@@ -0,0 +1,30 @@
...
+cd ..
+
+git clone https://github.com/tpm2-software/tpm2-totp
+cd tpm2-totp
+./bootstrap
+./configure
+make
+sudo make install
+
+sudo ldconfig
```
eventualbuddha, line 22: Not super self-explanatory.

setup-tpm2-totp.sh
```diff
@@ -0,0 +1,30 @@
...
+mkdir ./tpm2-totp
+cd tpm2-totp
+
+git clone https://github.com/tpm2-software/tpm2-tss
+cd tpm2-tss
+./bootstrap
+./configure
+make 
+sudo make install
+cd ..
```
eventualbuddha, line 13: Use sub-shell?

```suggestion
(
cd tpm2-tss
./bootstrap
./configure
make 
sudo make install
)
```

setup-tpm2-totp.sh
```diff
@@ -0,0 +1,30 @@
+
```
eventualbuddha, line 1:
```suggestion
set -euo pipefail

```

Should this be executable? The other `setup-*.sh` ones aren't, but…

setup-tpm2-totp.sh
umbernhard: basically the TPM2-TOTP module gets installed and has to be loaded before it can be used.

eventualbuddha: A comment to this effect would be nice is all I meant 😄

setup-tpm2-totp.sh
benadida: but we need them for the build right? So you're saying we can remove them afterwards?

setup-tpm2-totp.sh
benadida: I think I'm going to move to submodules, as per @eventualbuddha, and structure them under `tpm/` or the like. What do you think?

setup-tpm2-totp.sh
benadida: for now, I'm leaving this script as a separate thing to run manually before `setup-machine.sh`. But yes, when we have it as part of the default build, then yeah we'll have to pause and confirm that the QR code has been recorded.

config/sudoers
benadida: I don't see anything that would be a security hole, but you can reset the secret, which would then make all codes not match. I'm not too worried about it for now... unless there's a clear way to limit the parameters? Do we have to wrap it in a separate executable?

setup-tpm2-totp.sh
benadida: I think you're right about using submodules. Let's do that.

setup-tpm2-totp.sh
benadida: I will memorialize this as the one time @eventualbuddha asked me to add a comment, rather than take one away ;)

setup-tpm2-totp.sh
benadida: today I learned.

setup-tpm2-totp.sh
umbernhard: I don't think we do need them for build. The configure step will turn off the docs building parts of the build

setup-tpm2-totp.sh
umbernhard: that sounds pretty good to me. I would maybe say something like `tpm2-tools`or something like that.

umbernhard: (`tpm2-software` is the github org that produces both of these packages, so that's why I suggested it)

setup-tpm2-totp.sh
umbernhard: Ah, my bad I  didn't realize this was a separate script.

umbernhard (approved):
Looks good. I assume you confirmed that removing those four packages still cause a successful build? I wonder if we should figure out a build env for CirclCI or something...

benadida:
> Looks good. I assume you confirmed that removing those four packages still cause a successful build? I wonder if we should figure out a build env for CirclCI or something...

we don't have good tests for this package yet... we need to do this in a VM etc. In any case I am actively confirming that this works right now on real hardware :)

benadida:
@umbernhard note the additional commit I just added, which does a `clean` before creating the TOTP, so that this script can be run more than once on a given machine. I hit this case because the code survived an OS reinstall :) Kinda cool!

In the process, I've confirmed this works as is on real hardware, especially with the `clean` command added.

umbernhard:
@benadida yep that's a great idea!

eventualbuddha (approved):
This seems reasonable, but I couldn't actually get it working on my VM. I assume that's expected.

```
 /bin/mkdir -p '/usr/local/bin'
  /bin/bash ./libtool   --mode=install /usr/bin/install -c tpm2-totp '/usr/local/bin'
libtool: install: /usr/bin/install -c .libs/tpm2-totp /usr/local/bin/tpm2-totp
 /bin/mkdir -p '/usr/local/libexec/tpm2-totp'
  /bin/bash ./libtool   --mode=install /usr/bin/install -c plymouth-tpm2-totp '/usr/local/libexec/tpm2-totp'
libtool: install: /usr/bin/install -c .libs/plymouth-tpm2-totp /usr/local/libexec/tpm2-totp/plymouth-tpm2-totp
 /bin/mkdir -p '/usr/local/include'
 /usr/bin/install -c -m 644 include/tpm2-totp.h '/usr/local/include'
 /bin/mkdir -p '/usr/local/etc/initramfs-tools/hooks'
 /usr/bin/install -c dist/initramfs-tools/hooks/tpm2-totp '/usr/local/etc/initramfs-tools/hooks'
 /bin/mkdir -p '/usr/local/etc/initramfs-tools/scripts/init-premount'
 /usr/bin/install -c dist/initramfs-tools/scripts/init-premount/tpm2-totp '/usr/local/etc/initramfs-tools/scripts/init-premount'
 /bin/mkdir -p '/usr/local/lib/pkgconfig'
 /usr/bin/install -c -m 644 dist/tpm2-totp.pc '/usr/local/lib/pkgconfig'
make  install-data-hook
make[2]: Entering directory '/home/brian/src/vxsuite-complete-system/tpm2-software/tpm2-totp'
Makefile:2270: warning: overriding recipe for target 'dist-hook'
Makefile:2262: warning: ignoring old recipe for target 'dist-hook'
make[2]: Nothing to be done for 'install-data-hook'.
make[2]: Leaving directory '/home/brian/src/vxsuite-complete-system/tpm2-software/tpm2-totp'
make[1]: Leaving directory '/home/brian/src/vxsuite-complete-system/tpm2-software/tpm2-totp'
ERROR:tcti:src/tss2-tcti/tcti-device.c:443:Tss2_Tcti_Device_Init() Failed to open specified TCTI device file /dev/tpmrm0: No such file or directory 
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:154:tcti_from_file() Could not initialize TCTI file: libtss2-tcti-device.so.0 
ERROR:tcti:src/tss2-tcti/tcti-device.c:443:Tss2_Tcti_Device_Init() Failed to open specified TCTI device file /dev/tpm0: No such file or directory 
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:154:tcti_from_file() Could not initialize TCTI file: libtss2-tcti-device.so.0 
WARNING:tcti:src/util/io.c:253:socket_connect() Failed to connect to host 127.0.0.1, port 2321: errno 111: Connection refused 
ERROR:tcti:src/tss2-tcti/tcti-swtpm.c:591:Tss2_Tcti_Swtpm_Init() Cannot connect to swtpm TPM socket 
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:154:tcti_from_file() Could not initialize TCTI file: libtss2-tcti-swtpm.so.0 
WARNING:tcti:src/util/io.c:253:socket_connect() Failed to connect to host 127.0.0.1, port 2321: errno 111: Connection refused 
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:154:tcti_from_file() Could not initialize TCTI file: libtss2-tcti-mssim.so.0 
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:254:tctildr_get_default() No standard TCTI could be loaded 
ERROR:tcti:src/tss2-tcti/tctildr.c:428:Tss2_TctiLdr_Initialize_Ex() Failed to instantiate TCTI 
ERROR in main (src/tpm2-totp.c:425): 0xa000a - tcti:IO failure
```

benadida:
> This seems reasonable, but I couldn't actually get it working on my VM. I assume that's expected.

Correct, I have yet to be able to get a virtual TPM working in my VM, too. I have confirmed this on physical hardware.

That's one reason I'm keeping this script separate for now.

PR #88, opened 2021-09-14
Adds setup script for key generation, working towards #853

umbernhard:
Installs signify-openbsd and qrencode, creates a keypair and outputs a QR code corresponding to the public key.

### DIFF
--- a/setup-signify.sh
+++ b/setup-signify.sh
@@ -4,10 +4,10 @@ set -euo pipefail
 sudo apt -y install signify-openbsd qrencode
 
 # clear out old keys
-rm -f key.pub key.sec
+rm -f /vx/config/key.pub /vx/config/key.sec
 
 # Generate the keypair
 signify-openbsd -G -n -p /vx/config/key.pub -s /vx/config/key.sec
 
 # Output the public key for enrollment into another device
-qrencode -t UTF8 -r key.pub -o -
+qrencode -t UTF8 -r /vx/config/key.pub -o -
--- a/setup-signify.sh
+++ b/setup-signify.sh
@@ -1,13 +1,21 @@
 #!/bin/bash
 set -euo pipefail 
 
-sudo apt -y install signify-openbsd qrencode
+if [ $# -eq 1 ]; then
+    wd=$1
 
-# clear out old keys
-rm -f /vx/config/key.pub /vx/config/key.sec
+    sudo apt -y install signify-openbsd qrencode
 
-# Generate the keypair
-signify-openbsd -G -n -p /vx/config/key.pub -s /vx/config/key.sec
+    # clear out old keys
+    rm -f $wd/key.pub $wd/key.sec
 
-# Output the public key for enrollment into another device
-qrencode -t UTF8 -r /vx/config/key.pub -o -
+    # Generate the keypair
+    signify-openbsd -G -n -p $wd/key.pub -s $wd/key.sec
+
+    # Output the public key for enrollment into another device
+    qrencode -t UTF8 -r $wd/key.pub -o -
+else 
+    echo "Usage: ./setup-signify [directory]"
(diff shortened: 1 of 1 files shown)


setup-signify.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -euo pipefail 
+
+sudo apt -y install signify-openbsd qrencode
+
+# clear out old keys
+rm -f key.pub key.sec
```
benadida, line 7: should this be `/vx/config/key.pub` etc.?

setup-signify.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Generate the keypair
+signify-openbsd -G -n -p /vx/config/key.pub -s /vx/config/key.sec
+
+# Output the public key for enrollment into another device
+qrencode -t UTF8 -r key.pub -o -
```
benadida, line 13: same here, should this be `/vx/config/key.pub` ?

benadida (commented):
wondering about paths

setup-signify.sh
umbernhard: oh whoops, yeah.

setup-signify.sh
umbernhard: Fixed

setup-signify.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -euo pipefail 
+
+sudo apt -y install signify-openbsd qrencode
+
+# clear out old keys
+rm -f /vx/config/key.pub /vx/config/key.sec
+
+# Generate the keypair
+signify-openbsd -G -n -p /vx/config/key.pub -s /vx/config/key.sec
```
eventualbuddha, line 10: I'm unsure what the permissions should be for `/vx/config` to be able to run this without `sudo`.

eventualbuddha (approved):
Mostly okay, though unsure about a permission issue.

setup-signify.sh
umbernhard: There's no reason we couldn't run with sudo I don't think, @benadida is that right?

benadida: vx-admin can write to /Vx/config, and that's the user that will generate the keys.

umbernhard: so no need for sudo then?

setup-signify.sh
```diff
@@ -1,13 +1,21 @@
...
+    rm -f $wd/key.pub $wd/key.sec
+
+    # Generate the keypair
+    signify-openbsd -G -n -p $wd/key.pub -s $wd/key.sec
+
+    # Output the public key for enrollment into another device
+    qrencode -t UTF8 -r $wd/key.pub -o -
+else 
+    echo "Usage: ./setup-signify [directory]"
+    exit;
```
eventualbuddha, line 19:
```suggestion
    exit 1
```

setup-signify.sh
```diff
@@ -1,13 +1,21 @@
 #!/bin/bash
 set -euo pipefail 
 
-sudo apt -y install signify-openbsd qrencode
+if [ $# -eq 1 ]; then
+    wd=$1
```
eventualbuddha, line 5:
```suggestion
  public_path="${1}/key.pub"
  secret_path="${1}/key.sec"
```

Then use `"${public_path}"` and `"${secret_path}"` (with double-quotes) below.

benadida (approved):

setup-signify.sh
eventualbuddha: Or, just make this an early exit:

```bash
if [ $# -eq 0 ]; then
    echo "Usage: ./setup-signify [directory]"
    exit 1
fi

# put the rest here

eventualbuddha (approved):

PR #93, opened 2021-09-17
bump vxsuite and kiosk browser

benadida:


benadida:
merging this as it's just a bump, @eventualbuddha FYI.

PR #95, opened 2021-09-18
load signify secret key from config directory when running kiosk-browser

benadida:


benadida:
@eventualbuddha I'll merge and look for any post-hoc review comments.

PR #96, opened 2021-09-19
fix: run module smartcards when running bsd

benadida:


benadida:
@eventualbuddha @carolinemodic merging this small fix and look forward to any post-hoc comments.

PR #113, opened 2021-10-28
feat(vcs) log machine boots shutdowns and restarts

carolinemodic:
Adds logs for machine boots, shutdowns, and restarts. Will update type types in vxsuite to reflect these changes. 

A few things I'm open to feedback on

- I opted to reuse 1 event id for both powering down and rebooting, but the message in the log tells you which it is thats happening so we could make these unique event ids if that is preferred
- I'm very unsure about how granular to make the "event types" and what the distinctions should be, its not really necessary at all except its a required field in the CDF (but we get to define our own types with documentation for them) we could just make them message and error, for example. The distinction I tried to make here between system-action and system-status is that action is when something was initiated (like the boot/shutdown) and the status is just an update message about the status of some process. But ... I don't feel strongly. 
- I had to update rsyslog in order to get an easier way to create json logs, but its possible to do it in the default version of rsyslog with this ubuntu build if we don't want to do that, just uglier. 

Example logs after shutting down, booting, and then restarting:
`
{"timeLogWritten":"2021-10-27T18:20:03.146528-07:00", "host":"ubuntu", "source": "system", "eventId": "machine-shutdown-init", "eventType": "system-action", "user": "system", "disposition": "na", "message":"System is powering down."}
{"timeLogWritten":"2021-10-27T18:20:04.367379-07:00", "host":"ubuntu", "source": "system", "eventId": "machine-shutdown-complete", "eventType": "system-status", "user": "system", "disposition": "success", "message":"Shutting down."}
{"timeLogWritten":"2021-10-27T18:20:17.306429-07:00", "host":"ubuntu", "source": "system", "eventId": "machine-boot-init", "eventType": "system-action", "user": "system", "disposition": "na", "message": "Machine boot initiated"}
{"timeLogWritten":"2021-10-27T18:20:18.860397-07:00", "host":"ubuntu", "source": "system", "eventId": "machine-boot-complete", "eventType": "system-status", "user": "system", "disposition": "success", "message":"Startup finished in 4.007s (kernel) + 1.691s (userspace) = 5.698s."}
{"timeLogWritten":"2021-10-27T18:20:37.937462-07:00", "host":"ubuntu", "source": "system", "eventId": "machine-shutdown-init", "eventType": "system-action", "user": "system", "disposition": "na", "message":"System is rebooting."}
{"timeLogWritten":"2021-10-27T18:20:39.119368-07:00", "host":"ubuntu", "source": "system", "eventId": "machine-shutdown-complete", "eventType": "system-status", "user": "system", "disposition": "success", "message":"Shutting down."}
{"timeLogWritten":"2021-10-27T18:20:45.253301-07:00", "host":"ubuntu", "source": "system", "eventId": "machine-boot-init", "eventType": "system-action", "user": "system", "disposition": "na", "message": "Machine boot initiated"}
{"timeLogWritten":"2021-10-27T18:20:46.770967-07:00", "host":"ubuntu", "source": "system", "eventId": "machine-boot-complete", "eventType": "system-status", "user": "system", "disposition": "success", "message":"Startup finished in 3.828s (kernel) + 1.662s (userspace) = 5.490s."}`

### DIFF
--- a/config/30-votingworks.conf
+++ b/config/30-votingworks.conf
@@ -1,4 +1,4 @@
-template(name="outfmt" type="list" option.jsonf="on") {
+template(name="jsonfmt" type="list") {
          constant(value="{")
          property(outname="timeLogWritten" name="timereported" dateformat="rfc3339" format="jsonf")
          constant(value=",")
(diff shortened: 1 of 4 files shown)


config/30-votingworks.conf
```diff
@@ -9,7 +9,59 @@ template(name="outfmt" type="list" option.jsonf="on") {
  }
 
 # reformat json messages into one json blob and send to vx-logs.log
-if $syslogtag == "votingworksapp:" and re_match($msg,'^ *\\{.*\\} *$') then /var/log/vx-logs.log;outfmt
+if $syslogtag == "votingworksapp:" and re_match($msg,'^ *\\{.*\\} *$') then /var/log/vx-logs.log;jsonfmt
```
benadida, line 12: small preference for using `vxapp` instead of the lengthier `votingworksapp`, but not necessary to fix here, maybe something to consider separately?

setup-machine.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+!/bin/bash
```
benadida, line 1: what does this change do? I'm not used to seeing anything other than `#!`

benadida (approved):

setup-machine.sh
carolinemodic: oh ... I don't know how this got in here or why I did this lol, will revert

PR #119, opened 2021-12-01
build: update for new vxsuite layout

eventualbuddha:
Bumps vxsuite to a version that has a new repo layout, so this includes changes to account for it.

Closes votingworks/vxsuite#1273


carolinemodic (approved):
Yay! Thank you!

PR #125, opened 2022-01-07
feat(vcs) fix logging setup for ubuntu 20 and modularize setup script

carolinemodic:
Updates the complete-system logging setup for ubuntu 20, we no longer need to update rsyslog like we did on ubuntu 18. Also found some issues with the previous setup that I believe are unrelated to ubuntu 20 that I fixed. 

In order to make it easier to test the logging setup I moved the logging commands to their own setup script. This way you can get the logging configured such that the vx-logs.log file is being populated easily without fully locking down and setting up a machine. 

I moved the smaller setup scripts into their own folder to keep things more organized. It didn't seem like setup-tpm2-totp.sh and setup-signify.sh were referenced anywhere else so I wasn't sure if it was safe to move those.


setup-machine.sh
```diff
@@ -99,11 +94,10 @@ sudo rm -rf /vx/services/* /vx/ui/* /vx/admin/*
 
 # Let vx-admin read logs
 sudo usermod -aG adm vx-admin
+sudo usermod -aG adm vx-ui
```
carolinemodic, line 97: @benadida Does this add permissions other than just viewing the logs? Without this even with granting kiosk-browser read permissions to /var/log I would get a permissions denied error when trying to export the logs from the app.

benadida: yeah, I believe that is the key permission change, so I'm good with this.

benadida (approved):

eventualbuddha (approved):

PR #126, opened 2022-01-26
Update product names and BMD app mode options

jonahkagan:
Depends on https://github.com/votingworks/vxsuite/pull/1368

### DIFF
--- a/setup-machine.sh
+++ b/setup-machine.sh
@@ -23,7 +23,7 @@ MODEL_NAMES+=('VxAdmin')
 
 echo "${#CHOICES[@]}. Ballot Scanner"
 CHOICES+=('bsd')
-MODEL_NAMES+=('VxScan')
+MODEL_NAMES+=('VxCentralScan')
 
 echo "${#CHOICES[@]}. Ballot Marking Device (BMD)"
 CHOICES+=('bmd')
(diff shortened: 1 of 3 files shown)


setup-machine.sh
```diff
@@ -23,7 +23,7 @@ MODEL_NAMES+=('VxAdmin')
 
 echo "${#CHOICES[@]}. Ballot Scanner"
 CHOICES+=('bsd')
-MODEL_NAMES+=('VxScan')
+MODEL_NAMES+=('VxCentralScan')
 
 echo "${#CHOICES[@]}. Ballot Marking Device (BMD)"
```
jonahkagan, line 28: wondering if it matters that the bmd always has model name set to VxMark here

carolinemodic: When we configure the machines there is a way to get into an admin menu (tty2, you press, f2 or f3 or something...) where Drew (or whoever) can change the app mode, which also changes the model name. You can see that in the admin_functions scripts directory if you are curious.

carolinemodic: Oh maybe I don't actually change the model name I thought I did lol, regardless this environment variable isn't used for anything yet so its not a big deal.

carolinemodic (approved):

PR #130, opened 2022-02-02
Enhanced security

umbernhard:
This PR enables enhanced security features like dm-verity, a unified kernel, kernel command line,  and bootloader, and secure boot signing functionality. Note that these features only apply to debian builds of our system.


umbernhard:
Closes votingworks/vxsuite#1434 and votingworks/vxsuite#1437. Implements a secure boot enabled dm-verity system backed by TPM2-TOTP on Debian 11.

config/admin-functions/show-vx-suite-admin-menu.sh
```diff
@@ -22,6 +22,14 @@ while true; do
   echo -e "Machine Type: \e[32m${VX_MACHINE_TYPE}\e[0m"
   echo -e "Machine Manufacturer: \e[32m${VX_MACHINE_MANUFACTURER}\e[0m"
   echo -e "Machine Model Name: \e[32m${VX_MACHINE_MODEL_NAME}\e[0m"
+
+  # TODO: do we want to try to also dispaly secure boot status?
```
carolinemodic, line 26: nit: display*

carolinemodic (commented):
I'll let Ben sign off but the parts of this I understand seem reasonable to me. Am I reading this right that the setup flow will be that setup-machine does NOT set up dm-verity and then you have to go into tty2 and run lockdown for that to happen? 

Did you test that the non-secure boot stuff still works to setup a machine on ubuntu?

umbernhard:
I did test it on Ubuntu a while ago, but I will test it again to be sure. Should I test against 18, 20, or both?

config/admin-functions/lockdown.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Now do the dm-verity setup
+veritysetup format /dev/mapper/Vx--vg-root /dev/mapper/Vx--vg-hashes| tee "/tmp/verity.log"
+
+# Find the root hash and append it to our cmdline
+HASH="$(awk '/Root hash:/ { print $3 }' "/tmp/verity.log")"
+echo "$(cat /vx/admin/config/cmdline)${HASH}" > /tmp/cmdline
+
+# TODO: Make sure the output initramfs/kernel file names are correct!
+# Now package up ouer kernel, cmdline, etc
```
benadida, line 17:
```suggestion
# Now package up our kernel, cmdline, etc
```

config/admin-functions/lockdown.sh
```diff
@@ -0,0 +1,50 @@
...
+    --add-section .osrel="/usr/lib/os-release" --change-section-vma .osrel=0x20000 \
+    --add-section .cmdline="/tmp/cmdline" --change-section-vma .cmdline=0x30000 \
+    --add-section .splash="/vx/admin/config/logo.bmp" --change-section-vma .splash=0x40000 \
+    --add-section .linux="/boot/vmlinuz-5.10.0-11-amd64" --change-section-vma .linux=0x2000000 \
+    --add-section .initrd="/boot/initrd.img-5.10.0-11-amd64" --change-section-vma .initrd=0x3000000 \
+    "/usr/lib/systemd/boot/efi/linuxx64.efi.stub" "/tmp/linux.efi"
+
+# Sign the resulting binary
+# First make sure the drive is mounted
+mount /dev/sda /mnt
```
benadida, line 28: what happens if this fails cause forgot to insert USB? I can live with not fixing this yet, but next quick fix would be to try this first and if that fails, error out and say "insert USB stick please"

config/admin-functions/show-vx-suite-admin-menu.sh
```diff
@@ -118,9 +129,15 @@ while true; do
     
     resettotp)
         sudo tpm2-totp clean || true
-        sudo tpm2-totp --pcrs=0,7 init
+        sudo tpm2-totp --pcrs=0,2,4,5,7 init
```
benadida, line 132: 😍

setup-machine.sh
```diff
@@ -67,24 +73,44 @@ sudo apt install -y unclutter mingetty pmount brightnessctl
...
+sudo ln -sf /var/vx/data /vx/data
+
+echo "Creating users"
 # create users, no common group, specified uids.
-sudo useradd -u 750 -m -d /vx/services vx-services
-sudo useradd -u 751 -m -d /vx/ui -s /bin/bash vx-ui
-sudo useradd -u 752 -m -d /vx/admin -s /bin/bash vx-admin
+id -u vx-services &> /dev/null || sudo useradd -u 750 -m -d /var/vx/services vx-services
+id -u vx-ui &> /dev/null || sudo useradd -u 751 -m -d /var/vx/ui -s /bin/bash vx-ui
+id -u vx-admin &> /dev/null || sudo useradd -u 752 -m -d /var/vx/admin -s /bin/bash vx-admin
```
benadida, line 102: idempotence! YAY!

benadida (approved):
lgtm. I'm okay if we say that locking down is only done on Debian from now on. Big decision but we kind of have to make it, right?

config/admin-functions/lockdown.sh
umbernhard: this is a good question. I will test and fix accordingly.

umbernhard: So I tested with the USB unplugged, and with another USB plugged in that didn't have keys on it. When mount fails it just loops back into the admin interface. The only potentially sticky situation is if two USB block devices are plugged in at the same time, and the keys drive gets a name that isn't `sda` (like it could get `sdb`). Do we care?

benadida: Nope, that's good enough for me.

umbernhard:
I tested locking down on Ubuntu 18 and it seems like it worked just fine. I'm gonna merge.

PR #132, opened 2022-02-07
Update setup-speech-synthesis.sh

umbernhard:
Newer versions of speech-dispatcher and/or festival use a different path.


setup-scripts/setup-speech-synthesis.sh
```diff
@@ -13,16 +13,21 @@ sudo apt install -y festival speech-dispatcher speech-dispatcher-festival festvo
...
+	sudo mv cmu_us_slt_arctic /usr/share/festival/voices/us/cmu_us_slt_arctic_clunits
+fi
+  
 # done with downloaded files
 cd "$INSTALL_DIR"
 rm -rf /tmp/vx
 
 # set up festival voice
 sudo cp config/speechd.conf /etc/speech-dispatcher/
-sudo systemctl restart speech-dispatcher
```
carolinemodic, line 25: why are we removing this?

carolinemodic (approved):

setup-scripts/setup-speech-synthesis.sh
umbernhard: in the newer versions of speech-dispatcher, it doesn't rely on a systemd unit. It gets launched by a different part of X, so the systemd unit doesn't work. Worse, it's always inactive and cannot be started, which throws an error if you try to restart it.

PR #137, opened 2022-02-24
feat(vcs) fix printer driver for hp laserjet 404 on debian

carolinemodic:
I'm not sure if this ppd file will work on ubuntu so I opted to just reference a different printer config for debian and non-debian. I'm also guessing we will need to update the ppd file for the other HP printer as well but I don't have it so I can not easily test that. I'm not 100% sure that this driver does everything we need, I'm guessing do to its "generic" nature it might not support some of the printers features but ... it seems to print fine from my limited testing?


umbernhard (approved):
Looks good to me. I'm a little surprised that the Ubuntu drivers don't work, but on the other hand one of the things Ubuntu does is make a lot of that stuff "just work".

carolinemodic:
Tested IPP and it seems to still work.

PR #140, opened 2022-03-02
feat(vcs) allow vx-ui to run sudo efibootmgr

carolinemodic:
See https://github.com/votingworks/kiosk-browser/pull/105


umbernhard:
@benadida I imagine this dovetails with the stuff you've been working on with clonezilla, should we also add it for vx-admin?

umbernhard (approved):

benadida:
> @benadida I imagine this dovetails with the stuff you've been working on with clonezilla, should we also add it for vx-admin?

We could yes, or we could have a separate script like lockdown.sh but just the specific efibootmgr call, and give that permission to `vx-admin`. In any case, I'm comfortable doing it separately.

PR #144, opened 2022-03-04
Move efibootmgr command to separate script so it can be run on cloned devices. 

umbernhard:


config/admin-functions/lockdown.sh
```diff
@@ -36,13 +36,7 @@ OUTDIR="${EFIDIR}/${TARGET}"
 DEV="$(df "$OUTDIR" | tail -1 | cut -d' ' -f1)"
 part=$(cat /sys/class/block/$(basename $DEV)/partition)
```
benadida, line 37: gotta move these too, I think?

config/admin-functions/show-vx-suite-admin-menu.sh
```diff
@@ -138,6 +141,9 @@ while true; do
         read -s -n 1
     ;;
     
+    setupbootentry)
+    	bash setup-boot-entry.sh
+	read -s -n 1
```
benadida, line 147: missing `;;` I think?

benadida (changes requested):
caught a couple of bugs by inspection, I think, but check me. Make sure this runs properly independently and as part of the calling script.

config/admin-functions/lockdown.sh
umbernhard: ope, good call.

config/admin-functions/show-vx-suite-admin-menu.sh
umbernhard: also good call.

benadida (approved):

benadida:
testing this on a real machine, and seeing a couple of issues:

- the lockdown script fails on first run... but error message disappears quickly so I don't have any more info yet, I'll see what I can find.
- the setup-boot-entry script doesn't work because it doesn't have the executable bit set.

benadida:
I tried ctrl-c'ing out of the menu and doing `sudo ./lockdown.sh` at the prompt as `vx-admin`, and that worked on the first try. Maybe there's something wrong when it's run from the menu? I'll keep trying for a more helpful repro.

PR #146, opened 2022-03-10
Create INSTALL.md

umbernhard:
File will eventually provide instructions for provisioning production machines. Addresses votingworks/vxsuite#1534

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,121 @@
+This file explains getting VxSuite up and running in Debian 11.2, along with setting up security features like Secure Boot, dm-verity, and TPM2-TOTP. 
+
+<h2>Installer</h2>
+
+The install process can follow the usual path of the Debian installer. The only modification that needs to be made is as follows:
+
+Set the machine's hostname to be "VxMark" (TODO: make this more general):
+![image](https://user-images.githubusercontent.com/2686765/156217619-95165aca-da51-406d-8c93-4630a5e50a63.png)
+
+On the disk partitioning screen, select “Setup LVM”
+
+![image](https://user-images.githubusercontent.com/2686765/156217724-78ada600-fb7b-4b93-b9f1-6ac5d1bd3f0d.png)
+
+Select the disk in question:
+![image](https://user-images.githubusercontent.com/2686765/156218657-7e0a8327-6791-4f5e-a8d5-497e013671d3.png)
+
+Use separate partitions for `/home`, `/var`, and `/tmp`
+![image](https://user-images.githubusercontent.com/2686765/156218692-8b62b4dd-aa13-49ce-8553-38636c7b3970.png)
+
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -0,0 +1,121 @@
...
+![image](https://user-images.githubusercontent.com/2686765/156219210-ba23cf55-f0c2-4734-8a1c-6acb865fbc49.png)
+
+Now we're done!
+![image](https://user-images.githubusercontent.com/2686765/156219231-c051bb0f-a816-4dfe-871a-7e4edb6c4780.png)
+![image](https://user-images.githubusercontent.com/2686765/156219254-c00b5fe9-53c7-411d-88c2-c6cf798fa665.png)
+
+Note: there may be a screen asking if we want to install with a UEFI-based bootloader. Say yes. Afterwards, continue through the rest of the install as normal. 
+
+<h2>First boot</h2>
+Since Debian does not have the same packages as Ubuntu (i.e. no PPAs), some modifications are needed from the usual VxSuite build process. First, you’ll have to add your user to the sudoers file using /sbin/visudo
```
benadida, line 53: I would suggest

`usermod -a -G sudo ${USER}`

INSTALL.md
```diff
@@ -0,0 +1,121 @@
...
+Copy the keys to the EFI system partition:
+```bash
+cp *.auth /boot/efi
+```
+Now reboot the machine and enter the firmware interface. On the Secure Boot screen, there should be an option to "Enroll Keys". It should show you your ESP on selection, at which point you can navigate to each of the three keys and enroll them in the firmware. As above, the order matters, so do DB, KEK then PK in that order.
+
+At this point there should be keys enrolled in firmware and the machine should be in Secure Boot mode. You will probably have to turn it off for now so that we can still boot Debian, since we haven't signed anything yet. 
+
+**IMPORTANT**: We are not using disk encryption, meaning that keys which are left on `/etc/efi-keys` or `/boot/efi` are **exposed**. The keys **must** be wiped from the disk before putting the machines in the field, preferably before cloning. 
+
```
benadida, line 96: should we cut this out of here into a separate doc for key generation since we're now locking down based on a USB stick?

INSTALL.md
```diff
@@ -0,0 +1,121 @@
...
+
+<h2>Setting up VxSuite</h2>
+
+Now we're ready to setup VxSuite. After rebooting into the OS,
+
+```bash
+git clone git@github.com:votingworks/vxsuite-complete-system
+cd vxsuite-complete-system
+git checkout enhanced-security
```
benadida, line 105:
```suggestion
```

benadida (approved):
some tweaks

INSTALL.md
umbernhard: yes, yes we should.

PR #150, opened 2022-03-15
feat(admin-menu): Basic Configuration Wizard

jonahkagan:
Task: https://github.com/votingworks/vxsuite/issues/1448

- Add a basic configuration wizard for configuring the Machine ID, setting the clock, setting up TOTP, and generating signing keys
- Separate the admin menu into two sections, one for basic configuration containing just the wizard, and one for advanced actions (everything else)

To make the wizard work well, there are some tweaks to individual commands:
- Setting the Machine ID asks for confirmation
- Clock setting is more user friendly, prompting for each component of the date/time separately
- TOTP setup hides some error output when cleaning if TOTP hasn't been setup yet

### DIFF
--- a/config/admin-functions/basic-configuration.sh
+++ b/config/admin-functions/basic-configuration.sh
@@ -22,15 +22,15 @@ ${VX_FUNCTIONS_ROOT}/set-clock.sh
 echo
 echo -e "\e[1mStep 3: Record Machine Metadata\e[0m"
 echo 'Setting up TOTP...'
-TOTP_URI=`${VX_FUNCTIONS_ROOT}/reset-totp.sh | grep otpauth`
+TOTP_URI=$(${VX_FUNCTIONS_ROOT}/reset-totp.sh | grep otpauth)
 echo "TOTP URI: ${TOTP_URI}"
 echo 'Setting up signing keys...'
 ${VX_FUNCTIONS_ROOT}/generate-key.sh > /dev/null
-PUBLIC_KEY=`cat "${VX_CONFIG_ROOT}/key.pub" | tail -n 1`
+PUBLIC_KEY=$(cat "${VX_CONFIG_ROOT}/key.pub" | tail -n 1)
 echo "Public Signing Key: ${PUBLIC_KEY}"
 echo "Record this QR code containing the Machine ID, TOTP URI, and Public Signing Key:"
-MACHINE_ID=`cat "${VX_CONFIG_ROOT}/machine-id"`
-echo -e "${MACHINE_ID}\n${TOTP_URI}\n${PUBLIC_KEY}" | qrencode -t ANSIUTF8 -o -
+MACHINE_ID=$(< "${VX_CONFIG_ROOT}/machine-id")
+echo -e "${MACHINE_ID}\n${TOTP_URI}\n${PUBLIC_KEY}" | qrencode -t ANSI -o -
 
 while true; do
     read -p "Confirm QR code recorded (y/n) " CONFIRM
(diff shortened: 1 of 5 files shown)


config/admin-functions/basic-configuration.sh
```diff
@@ -0,0 +1,43 @@
...
+${VX_FUNCTIONS_ROOT}/choose-vx-machine-id.sh
+
+echo
+echo -e "\e[1mStep 2: Set Clock\e[0m"
+${VX_FUNCTIONS_ROOT}/set-clock.sh
+
+echo
+echo -e "\e[1mStep 3: Record Machine Metadata\e[0m"
+echo 'Setting up TOTP...'
+TOTP_URI=`${VX_FUNCTIONS_ROOT}/reset-totp.sh | grep otpauth`
```
eventualbuddha, line 25: I generally prefer `$(…)` over backticks because it's possible to nest `$(…)`. Up to you, though.

config/admin-functions/basic-configuration.sh
```diff
@@ -0,0 +1,43 @@
...
+echo "TOTP URI: ${TOTP_URI}"
+echo 'Setting up signing keys...'
+${VX_FUNCTIONS_ROOT}/generate-key.sh > /dev/null
+PUBLIC_KEY=`cat "${VX_CONFIG_ROOT}/key.pub" | tail -n 1`
+echo "Public Signing Key: ${PUBLIC_KEY}"
+echo "Record this QR code containing the Machine ID, TOTP URI, and Public Signing Key:"
+MACHINE_ID=`cat "${VX_CONFIG_ROOT}/machine-id"`
```
eventualbuddha, line 32: `read-vx-machine-config.sh` does this, if you want to be consistent: `"$(< "${VX_CONFIG_ROOT}/machine-id")"`.

config/admin-functions/reset-totp.sh
```diff
@@ -2,5 +2,5 @@
 
 set -euo pipefail
 
-sudo tpm2-totp clean || true
+sudo tpm2-totp clean 2>/dev/null || true
```
eventualbuddha, line 5: Why?

config/admin-functions/set-clock.sh
```diff
@@ -2,36 +2,61 @@
...
 
+echo "Let's set the time"
 while true; do
-    read -p "Timezone - (P)acific, (M)ountain, (C)entral, (E)astern: " TZ
-    if [[ "${TZ}" =~ ^[P,M,C,E]$ ]]; then
-	break
-    fi
+    while true; do
+        read -p "Pick a timezone - (p)acific, (m)ountain, (c)entral, (e)astern: " TZ
+        [[ "${TZ}" =~ ^p|m|c|e$ ]] && break
```
eventualbuddha, line 45: Case-insensitive match:
```suggestion
        [[ "${TZ,,}" =~ ^p|m|c|e$ ]] && break
```

eventualbuddha (approved):

config/admin-functions/reset-totp.sh
jonahkagan: If totp hasn't been initialized, the `clean` command prints an error saying you need to initialize it first. The `|| true` here handles that case by ignoring the error exit code, but I also wanted to suppress the output.

config/admin-functions/set-clock.sh
jonahkagan: I think it's fine to be case-sensitive, and a bit simpler

PR #151, opened 2022-03-15
fix(build): Handle build errors correctly

jonahkagan:
Due to some quirks of bash's `set -e` command and how it interacts with subshells, errors in the `build.sh` script were not getting handled correctly. The script would continue executing after a build error and print a success message.

More here: https://unix.stackexchange.com/questions/65532/why-does-set-e-not-work-inside-subshells-with-parenthesis-followed-by-an-or

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -63,7 +63,6 @@ build() {
     cd "${DIR}/vxsuite/frontends/${APP}"
     pnpm install
     BUILD_ROOT="${BUILD_ROOT}/vxsuite" ./script/prod-build
-    echo $?
 
     cp -rp \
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -59,6 +63,7 @@ build() {
     cd "${DIR}/vxsuite/frontends/${APP}"
     pnpm install
     BUILD_ROOT="${BUILD_ROOT}/vxsuite" ./script/prod-build
+    echo $?
```
eventualbuddha, line 66:
```suggestion
```

eventualbuddha (approved):

PR #153, opened 2022-03-16
fix(setup-machine): Add sudo to timedatectl command

jonahkagan:
That way we don't get stuck waiting for a password on this step when running in dev.


jonahkagan:
Closing in favor of #154

jonahkagan:
Reopening while #154 is getting tested

carolinemodic (approved):

PR #156, opened 2022-03-16
Update boot entry path

umbernhard:
that way the lockdown script will actually work :sweat_smile:

### DIFF
--- a/config/admin-functions/lockdown.sh
+++ b/config/admin-functions/lockdown.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -euo pipefail 
 
+: "${VX_FUNCTIONS_ROOT:="$(dirname "$0")"}"
+
 update-initramfs -u
 
 # Remount / so it can't change while we're doing the veritysetup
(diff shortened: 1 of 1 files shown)


config/admin-functions/lockdown.sh
```diff
@@ -30,7 +30,7 @@ mount /dev/sda /mnt || mount /dev/sda1 /mnt || (echo "Secure boot keys not found
 sbsign --key=/mnt/DB.key --cert=/mnt/DB.crt --output /boot/efi/EFI/debian/VxLinux-signed.efi /tmp/linux.efi
 
 # Now install it 
-bash setup-boot-entry.sh
+bash /vx/admin/admin-functions/setup-boot-entry.sh
```
jonahkagan, line 33: I noticed that a lot of the admin-functions scripts pull out these paths into variables. E.g. at the top of `show-vx-suite-admin-menu.sh`:

```bash
: "${VX_FUNCTIONS_ROOT:="$(dirname "$0")"}"
: "${VX_CONFIG_ROOT:="/vx/config"}"
```

Do you think something like that would be appropriate here?

benadida: Yes, +1, we should do the `VX_FUNCTIONS_ROOT` thing that defaults to `dirname "$0"`

benadida (changes requested):
path thing

benadida (approved):

PR #159, opened 2022-03-18
feat: Run basic configuration on first boot

jonahkagan:
Task: https://github.com/votingworks/vxsuite/issues/1448

- In setup-machine.sh, set a file flag RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT
- Configure an autologin script that checks for this flag. If the flag is present, it logs in vx-admin. Otherwise, it logs in vx-ui like normal
- When vx-admin is logged in, always run the admin menu script (regardless of which tty we're in, since in this flow it will be tty1 instead of the usual tty2)
- When showing the admin menu, if we see the file flag is set, just run the basic configuration wizard
- At the end of the wizard, remove the file flag
- Also adds an option to the menu to remove the file flag at any time

### DIFF
--- a/config/admin-functions/basic-configuration.sh
+++ b/config/admin-functions/basic-configuration.sh
@@ -37,6 +37,10 @@ while true; do
     [[ "${CONFIRM}" = "y" ]] && break
 done
 
+if [[ -f "${VX_CONFIG_ROOT}/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT" ]]; then
+    rm -f "${VX_CONFIG_ROOT}/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT"
+fi
+
 echo
 echo -e "\e[1mBasic Configuration Complete\e[0m"
 read -p "You must reboot for these changes to take effect. Reboot now? (y/n) " CONFIRM
--- a/config/admin-functions/show-vx-suite-admin-menu.sh
+++ b/config/admin-functions/show-vx-suite-admin-menu.sh
@@ -5,6 +5,11 @@ set -euo pipefail
 : "${VX_FUNCTIONS_ROOT:="$(dirname "$0")"}"
 : "${VX_CONFIG_ROOT:="/vx/config"}"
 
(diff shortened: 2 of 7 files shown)


config/admin-functions/reset-totp.sh
```diff
@@ -2,5 +2,6 @@
 
 set -euo pipefail
 
-sudo tpm2-totp clean 2>/dev/null || true
-sudo tpm2-totp --pcrs=0,2,4,5,7 init
\ No newline at end of file
+# sudo tpm2-totp clean 2>/dev/null || true
+# sudo tpm2-totp --pcrs=0,2,4,5,7 init
+echo otpauth://fake-url
```
benadida: I imagine this is for dev

setup-machine.sh
```diff
@@ -282,11 +282,15 @@ sudo chmod -R u=rwX /var/vx/config
...
 
+# set the flag to run basic configuration when the machine boots
+sudo touch /var/vx/config/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT
+
 # non-graphical login
 sudo systemctl set-default multi-user.target
 
 # setup auto login
 sudo mkdir -p /etc/systemd/system/getty@tty1.service.d
+sudo ln -s /vx/code/config/autologin.sh /vx/config/autologin.sh
```
benadida, line 293: why this? Can't we just run it directly from `/vx/code/config/autologin.sh` ?

benadida (approved):
overall looks good! Couple of small things.

setup-machine.sh
jonahkagan: it seemed like this was the pattern all throughout the setup-machine script, so I was just following it. also i saw a comment at the top that said /vx/code was read only - dunno if that's true.

config/admin-functions/reset-totp.sh
jonahkagan: removed!

setup-machine.sh
benadida: yes, it is read-only *after* we lock down the machine, which is why I think we want to put that script there, so it can't be modified later.

benadida:
tweaks I made:
- leave `autologin.sh` in the read-only `/vx/code/config` location
- only run the basic wizard on tty1, cause you might want the full menu on tty2 even if the flag is set
- don't set the flag in setup-machine, because a first-run in a VM won't want the wizard.

PR #160, opened 2022-03-18
Removes plymouth from boot

umbernhard:
This lets us show our custom BMP on boot.


umbernhard:
This issue closes votingworks/vxsuite#1569.

benadida (approved):

PR #167, opened 2022-03-23
fixes for read-only partition

benadida:
- remove setting the hostname with machine ID
- fix for timezone setting (confirms works for vx-admin, and with latest `kiosk-browser`, works for vx-ui, too.)

### DIFF
--- a/config/admin-functions/timedatectl
+++ b/config/admin-functions/timedatectl
@@ -3,9 +3,9 @@
 ORIGINAL_TIMEDATECTL=/usr/bin/timedatectl
 
 # override just the timezone setting part
-if [[ ${1} = "set-timezone" ]]; then
+if [[ "${1}" = "set-timezone" && -n "${2}" ]]; then
     TIMEZONE_LOCATION=`readlink /etc/localtime` || TIMEZONE_LOCATION="/etc/localtime"
-    ln -sf /usr/share/zoneinfo/$2 ${TIMEZONE_LOCATION}
+    ln -sf "/usr/share/zoneinfo/${2}" "${TIMEZONE_LOCATION}"
 else
     ${ORIGINAL_TIMEDATECTL} "$@"
 fi 
--- a/config/admin-functions/timedatectl
+++ b/config/admin-functions/timedatectl
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+ORIGINAL_TIMEDATECTL=/usr/bin/timedatectl
+
+# override just the timezone setting part
+if [[ ${1} = "set-timezone" ]]; then
+    TIMEZONE_LOCATION=`readlink /etc/localtime` || TIMEZONE_LOCATION="/etc/localtime"
+    ln -sf /usr/share/zoneinfo/$2 ${TIMEZONE_LOCATION}
+else
+    ${ORIGINAL_TIMEDATECTL} "$@"
+fi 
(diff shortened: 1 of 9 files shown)


config/admin-functions/timedatectl
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+ORIGINAL_TIMEDATECTL=/usr/bin/timedatectl
+
+# override just the timezone setting part
+if [[ ${1} = "set-timezone" ]]; then
+    TIMEZONE_LOCATION=`readlink /etc/localtime` || TIMEZONE_LOCATION="/etc/localtime"
+    ln -sf /usr/share/zoneinfo/$2 ${TIMEZONE_LOCATION}
```
eventualbuddha, line 8: Making this a little more defensive.
```suggestion
    ln -sf "/usr/share/zoneinfo/${2}" "${TIMEZONE_LOCATION}"
```

config/admin-functions/timedatectl
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+ORIGINAL_TIMEDATECTL=/usr/bin/timedatectl
+
+# override just the timezone setting part
+if [[ ${1} = "set-timezone" ]]; then
```
eventualbuddha, line 6: Sanity checking.
```suggestion
if [[ "${1}" = "set-timezone" && -n "${2}" ]]; then
```

config/admin-functions/hostnamectl
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+#
+# we set the hostname manually, rather than use hostnamectl,
+# so as to respect potential symlinks of the config files,
+# which is important for our locked down machine with read-only partitions.
+
+ORIGINAL_HOSTNAMECTL=/usr/bin/hostnamectl
+
+# override just the timezone setting part
```
eventualbuddha, line 9:
```suggestion
# override just the hostname setting part
```

config/admin-functions/hostnamectl
```diff
@@ -0,0 +1,19 @@
...
+if [[ ${1} = "set-hostname" ]]; then
```
eventualbuddha, line 10: Sanity checking.
```suggestion
if [[ "${1}" = "set-hostname" && -n "${2}" ]]; then
```

config/admin-functions/timedatectl
```diff
@@ -0,0 +1,11 @@
...
+    TIMEZONE_LOCATION=`readlink /etc/localtime` || TIMEZONE_LOCATION="/etc/localtime"
```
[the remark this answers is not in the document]
benadida: I don't think I want to do this, cause I only want to go *one* level down the symlink list, not all the way.

benadida:
tested in VM, working great. /cc @carolinemodic @eventualbuddha for a review whenever you can :)

config/admin-functions/timedatectl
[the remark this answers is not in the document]
eventualbuddha: Ah, I see. `/etc/localtime` is normally a symlink to e.g. `/usr/share/zoneinfo/America/Los_Angeles`, but we make it a symlink to `/var/etc/localtime` which is then a symlink to a `zoneinfo` path. So you're right, we don't want to traverse all the way.

That said, I don't think we need the hedge for the `readlink` failing, do we? It's always going to be a symlink, whether or not we've added our indirection. In the event we _haven't_ added the indirection, we'll actually be replacing e.g. `/usr/share/zoneinfo/America/Los_Angeles` with a link to itself which is … not great. Perhaps we should check that `TIMEZONE_LOCATION` is outside `/etc/share/zoneinfo`?

eventualbuddha (approved):

PR #168, opened 2022-03-23
Experimenting with separate /media partition

umbernhard:


eventualbuddha (approved):

benadida:
I think we can close this right @umbernhard ?

umbernhard:
Yep, good call.

PR #170, opened 2022-03-24
fix(preseed): larger root partition for vxscan

benadida:


eventualbuddha (approved):

umbernhard:
Do we just want to have a separate preseed file for VxScan?

benadida:
> Do we just want to have a separate preseed file for VxScan?

I would recommend same one for everything. Wasting 1-2GB in exchange for consistency / simplicity feels like the right tradeoff.

benadida:
confirmed that this works, changing to non-draft.

umbernhard (approved):

umbernhard:
hmm, I'm still getting an out of room error even with the 8G partition.

umbernhard:
this time it's `/home`

umbernhard:
oops, github shortcuts strike again. Sorry for the label!

umbernhard:
yep, works for me.

benadida:
> ah, I think it's because I had all the plustek stuff on my home partition. Lemme try removing that...

Yes, I hit that too... might be worth thinking about building out of `/var` and then clearing that out, rather than using `/home`, and then we can shrink `/home` even more. Next time!

PR #174, opened 2022-05-03
feat(vcs) add initial scripts for vxdev

carolinemodic:
Adds the initial scripts and setup for VxDev. A quick overview:

Setup-vxdev-machine.sh - Run once by Caroline when first configuring a new image for VxDev, should not need to be run again by the end user, a subset of the setup-machine script for production. There are a few things here that don't totally work yet like setting up the vx-admin user (I want to get the tty2 experience working similar to how it does in prod, etc.) 

Update-vxdev.sh - Configures VxDev to be for a specific machine type AND updates VxDev. There is some inception here where it checks out main and copies the VxDev scripts so that if we make changes we don't need people to reimage there machines. Those pieces are commented out in Roe's initial image so that it doesn't break when this doesn't exist on main yet. 

Update-code.sh - Asks the user whether they want to update to the latest code or to the latest tagged release and updates vxsuite and kiosk-browser to that, then builds. 

Run-VxWhatever.desktop - Desktop icon to actually run VxMark/VxScan/etc.


vxdev/run-vxadmin.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Version=1.0
+Name=VxAdmin
+Icon=votingworks-logo.png
+Exec=gnome-terminal -- /bin/sh -c 'env PATH=$PATH:/sbin /vx/code/vxsuite-complete-system/run.sh election-manager'
```
umbernhard, line 5: is the `gnome-terminal` call needed here? I would think just `/bin/sh -c` or `/bin/bash -c` would suffice.

vxdev/setup-vxdev-machine.sh
```diff
@@ -0,0 +1,183 @@
...
+
+# Debian doesn't add /sbin/ to default path, but it's needed for groupadd and other commands.
+# If /sbin/ is already in the path, this won't hurt, and it's just for running this script.
+if [[ $DISTRO == "Debian" ]]; then
+    export PATH=${PATH}:/sbin/
+fi
+
+if ! which plustekctl >/dev/null 2>&1
+then
+    echo "error: plustekctl was not found in PATH=${PATH}" >&2
```
umbernhard, line 19: This is not a change we need for this PR, but it sure would be nice to automate this instead of hoping the user pulled down plustekctl before running the script.

vxdev/setup-vxdev-machine.sh
```diff
@@ -0,0 +1,183 @@
...
+sudo bash setup-scripts/setup-signify.sh
+
+# turn off time synchronization
+sudo timedatectl set-ntp no
+
+# Install app to configure VxDev
+sudo mkdir -p /vx/scripts
+sudo cp vxdev/update-vxdev.sh /vx/scripts/.
+sudo cp vxdev/update-vxdev.desktop /usr/share/applications/.
+gsettings set org.gnome.shell favorite-apps "['update-vxdev.desktop','firefox-esr.desktop', 'org.gnome.Nautilus.desktop', 'org.gnome.Terminal.desktop', 'gnome-control-center.desktop']"
```
umbernhard, line 171: do we want Firefox as the default, or should we be grabbing chromium so it can be used to debug if kiosk-browser breaks?

vxdev/update-code.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+
+: "${VX_CONFIG_ROOT:="/vx/config"}"
+APP_TYPE=$(sudo cat "$VX_CONFIG_ROOT/machine-type")
+
+cd /vx/code/vxsuite-complete-system
+git checkout main > /dev/null
```
umbernhard, line 8: Is there any chance we'd want to pull down a branch that isn't main for pre-merge testing?

vxdev/update-code.sh
```diff
@@ -0,0 +1,63 @@
...
+	cd ..
+elif [[ $BRANCH == 'stable' ]]; then
+	git checkout $LATEST_TAG
+	git submodule foreach --recursive sudo git clean -xfd
+	git submodule update --init --recursive
+fi
+
+make build-kiosk-browser
+echo $APP_TYPE
+if [[ $APP_TYPE == 'VxCentralScan' ]] || [[ $APP_TYPE == 'VxAdminCentralScan' ]]; then
```
umbernhard, line 50: This is another not-for-this-PR suggestion, but we should maybe make `build.sh` accept environment variables in addition to command line flags, to simplify this kind of thing.

umbernhard (approved):
Looks good to me, just a few nitpicks here and there.

vxdev/update-code.sh
carolinemodic: Nothing that happens at this point relies on staying on the branch with the vxdev scripts as those are copied into other places on the drive earlier on. For the eng team internally people will definitely want to test things on branches sometimes but they can just go into vxsuite and go to the branch they want on their own for that and not run this program. At least initially.

vxdev/run-vxadmin.desktop
carolinemodic: I was having issues getting the desktop icon to work without launching it from a terminal, and the terminal was nice for debugging when there were errors but I will try again without this and make sure its necessary.

umbernhard: oh, one more thought: setting the PATH should also be unnecessary, since it's globally set by the setup script.

carolinemodic: Yes, I don't know why I have to do this but for some reason I do when it launches from the desktop icon it doesn't have sbin anymore 🤷‍♀️

umbernhard: huh, weird. I wonder if that would be true after a reboot?

benadida: I wonder if this has to do with where you're setting your `$PATH` environment variable. Is it in your `.bash_profile` ? Somewhere else?

benadida (approved):
lgtm

only one modest concern: it looks like there's copied code between locking down the machine and creating a vxdev machine... should that be factored out? Not a blocker, but a consideration for improvement for later.

PR #176, opened 2022-05-11
fix git output from showing up in terminal

carolinemodic:
removes some extraneous git command output from showing up when running the update and config programs in vxdev 

I also noticed I was sometimes getting weird permission errors when jumping back and forth between the latest code and the last stable release. Adding sudo fixed them but I admittedly don't really understand why there were issues and only on creating one specific file when there are multiple new files that have been added to vxsuite. But adding sudo felt safe.

### DIFF
--- a/vxdev/update-code.sh
+++ b/vxdev/update-code.sh
@@ -5,7 +5,7 @@
 APP_TYPE=$(sudo cat "$VX_CONFIG_ROOT/machine-type")
 
 cd /vx/code/vxsuite-complete-system
-git checkout main > /dev/null
+git checkout main > /dev/null 2>&1
 git pull > /dev/null
 git fetch --tags > /dev/null
 sudo git clean -xfd > /dev/null
--- a/vxdev/update-vxdev.sh
+++ b/vxdev/update-vxdev.sh
@@ -36,8 +36,8 @@ CHOICE=${CHOICES[$CHOICE_INDEX]}
 cd /vx/code/vxsuite-complete-system
 mkdir -p /vx/scripts
 
-git checkout main
-git pull
+git checkout main > /dev/null 2>&1
+git pull > /dev/null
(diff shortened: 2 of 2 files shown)


umbernhard:
Can you say more about the permissions errors you were seeing? It's probably not a great idea to be running git with sudo, it's definitely a security risk and it might cause future permissions problems.

carolinemodic:
@umbernhard There was a file that was added to vxsuite between m13b and now that I was getting a "permission denied could not create file" error and then my git status would have that as a locally deleted file and it wouldn't exist. There are other files I'm sure that have been added between m13b and now though so I don't understand why it was only that one, it is possible i did something weird in my VM state that wouldn't be a problem in general. After running `sudo git checkout` once and then running `git pull` without sudo I then got permission denied errors to check out a branch I think?

umbernhard:
I think it might be worth trying this on a clean install to see if we can repro (I can take a look at that tomorrow). It's _possible_ that there's something in the ESI refactor that may have affected this, but it seems unlikely. It's especially weird that it's only like one file. 

Running `sudo git checkout` could totally create permission problems if it updates the `.git` repo and sets its owner as root or something, I don't know for sure but I can play with it more tomorrow.

carolinemodic:
I have VxDev on some of my hardware so I can try to update those to the latest code and see if they have the problem tomorrow

benadida:
+1 to figuring out how to do this without the extra `sudo`, it probably means something is borked in the local git repo.

carolinemodic:
Ok I'll remove that commit, when I tested on my hardware VxDev setups I didn't repro this issue so I must have screwed up a permission in a weird way in my VM. I'm not convinced that its not going to happen that people can screw it up themselves somehow since I don't understand how it got into that state but we can cross the bridge if people see issues.

umbernhard (approved):

PR #189, opened 2022-05-12
Don't show lock screen in vxdev

carolinemodic:
The image will also need to be configured to automatically log the vx user in on boot


umbernhard (approved):
FWIW we can also set this in the preseed file I think.

PR #190, opened 2022-05-12
update to the right version of node before compiling the code

carolinemodic:
Updates to the right version of node (and any other vxsuite dependencies that get added to the setup script) before compiling the code

### DIFF
--- a/vxdev/update-code.sh
+++ b/vxdev/update-code.sh
@@ -30,11 +30,13 @@ then
     exit 1
 fi
 BRANCH=${CHOICES[$CHOICE_INDEX]}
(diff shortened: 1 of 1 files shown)


benadida (approved):

vxdev/update-code.sh
```diff
@@ -43,6 +45,9 @@ elif [[ $BRANCH == 'stable' ]]; then
 	git checkout $LATEST_TAG
 	git submodule foreach --recursive sudo git clean -xfd
 	git submodule update --init --recursive
+	cd vxsuite
+	./script/setup-dev
+	cd ..
```
eventualbuddha, line 50: FYI, this also works: `(cd vxsuite && ./script/setup-dev)`.

PR #199, opened 2022-07-01
add location for admin service data

benadida:


vxdev/run-vxadmin.desktop
```diff
@@ -2,6 +2,6 @@
 Version=1.0
 Name=VxAdmin
 Icon=runprogram.png
-Exec=gnome-terminal -- /bin/sh -c 'env PATH=$PATH:/sbin /vx/code/vxsuite-complete-system/run.sh election-manager'
+Exec=gnome-terminal -- /bin/sh -c 'env PATH=$PATH:/sbin env ADMIN_WORKSPACE=/vx/data/admin-service /vx/code/vxsuite-complete-system/run.sh election-manager'
```
carolinemodic, line 5: impressed you found this one haha

carolinemodic (approved):
I think you still have to manually set it with the default ./run.sh script though (I think thats how SCAN_WORKSPACE works too but at least now it will tell you that)

vxdev/run-vxadmin.desktop
benadida: > I think you still have to manually set it with the default ./run.sh script though (I think thats how SCAN_WORKSPACE works too but at least now it will tell you that)

Yes, same as SCAN_WORKSPACE, thus why I added the warning if it's not set.

benadida:
> I think you still have to manually set it with the default ./run.sh script though (I think thats how SCAN_WORKSPACE works too but at least now it will tell you that)

Yes, made it the same behavior as `SCAN_WORKSPACE`

PR #207, opened 2022-07-12
Switching from signify keys to tpm-based keys

umbernhard:
This PR swaps out signify keys to tpm-based keys. 

I've tested this on VxAdmin, but not much else atm.

One question: the current PCR policy I'm using locks the keys to a specific kernel version. We could relax that constraint by using PCR-policy signing, or just dropping the kernel measurement PCR, but then we might be vulnerable to rollback attacks. The alternative is to just generate new keys every time we do a software update that modifies the kernel. 

I think in principle just locking the policy to PCR 0 (the TPM's unique ID) and PCR 7 (the Secure Boot policy) may give us enough of what we want, but I'm curious to know @benadida's thoughts.

### DIFF
--- a/config/admin-functions/generate-key.sh
+++ b/config/admin-functions/generate-key.sh
@@ -4,10 +4,29 @@ set -euo pipefail
 
 : "${VX_CONFIG_ROOT:="/vx/config"}"
 
+
+set -euo pipefail 
+tpm2_startauthsession -S session.ctx
+tpm2_policypcr -S session.ctx -l "sha256:0,2,4,5,7" --policy pcr.policy
+tpm2_createprimary -C o -c primary.ctx
+
+# Save the primary context for reuse after reboot
+# First make sure nothing is in the handle we want to save our primary context
+# to
+tpm2_evictcontrol -Q -c 0x81000000 || true 2>&1 /dev/null
+tpm2_evictcontrol -c primary.ctx 0x81000000
+
+# Create a policy
+tpm2_create -L pcr.policy -u key.pub -C primary.ctx -G ecc -c key.ctx
+
+# Save the keys
+tpm2_evictcontrol -Q -c 0x81000001 || true 2>&1 /dev/null
+tpm2_evictcontrol -c key.ctx 0x81000001
+
+
 rm -f "${VX_CONFIG_ROOT}/key.pub" "${VX_CONFIG_ROOT}/key.sec"
-signify-openbsd -G -n -p "${VX_CONFIG_ROOT}/key.pub" -s "${VX_CONFIG_ROOT}/key.sec"
+tpm2_readpublic -c key.ctx -f PEM -o "${VX_CONFIG_ROOT}/key.pub"
+
 # Make the signing key readable by vx-group
 # We may want to further limit this in the future
-chgrp vx-group "${VX_CONFIG_ROOT}/key.sec"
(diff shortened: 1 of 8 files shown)


umbernhard:
For @benadida: to sign a blob, 

```
tpm2 startauthsession -S session.ctx --policy-session
tpm2 policypcr -S session.ctx -l "sha256:0,2,4,5,7" --policy pcr.policy
tpm2 sign -p session:session.ctx -c 0x81000001 -f plain -o signature.file data.blob
```

We can probably do some futzing around to make that pipeable, but it appears that tpm2-tools doesn't support input/output on stdin by default. You can make input work with a here-string:

```tpm2 sign -p session:session.ctx -c 0x81000001 -f plain -o signature.file <<< $(cat data_file)```

But I'm not sure how to handle output.

File: config/admin-functions/sign.sh [hunk 1]
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -euo pipefail
+
+tpm2 startauthsession -Q -S session.ctx --policy-session
+tpm2_policypcr -Q -S session.ctx -l "sha256:0,2,4,5,7" --policy pcr.policy
```
benadida, line 6: did we want this to be `0,7` and then differ in case of Surface Go?

carolinemodic (commented):
I will defer to Ben on this

config/admin-functions/sign.sh
benadida: @umbernhard just waiting for a fix for this before approving.

config/admin-functions/generate-key.sh
```diff
@@ -4,10 +4,29 @@ set -euo pipefail
...
-signify-openbsd -G -n -p "${VX_CONFIG_ROOT}/key.pub" -s "${VX_CONFIG_ROOT}/key.sec"
+tpm2_readpublic -c key.ctx -f PEM -o "${VX_CONFIG_ROOT}/key.pub"
+
 # Make the signing key readable by vx-group
 # We may want to further limit this in the future
-chgrp vx-group "${VX_CONFIG_ROOT}/key.sec"
-chmod g+r "${VX_CONFIG_ROOT}/key.sec"
-cat "${VX_CONFIG_ROOT}/key.pub" | qrencode -t ANSI -o -
\ No newline at end of file
+cat "${VX_CONFIG_ROOT}/key.pub" | qrencode -t ANSI -o -
```
benadida, line 32: the comment above this line is no longer relevant regarding public key readable by `vx-group`.

config/admin-functions/sign.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -euo pipefail
+
+tpm2 startauthsession -Q -S session.ctx --policy-session
+tpm2_policypcr -Q -S session.ctx -l "sha256:0,7" --policy pcr.policy
+
+# This expects the script to be run with the data to be signed passed in via 
+# a here-string, e.g. sudo ./sign.sh <<< "a string"
+# It "returns" the signature on stdout by catting a temporary file. 
+tpm2 sign -Q -p session:session.ctx -c 0x81000001 -f plain -o /tmp/sig
```
benadida, line 11: we should use a random filename, not a fixed filename, or we will get caught with a race condition later that will drive us up the wall.

benadida (approved):
couple small comments, otherwise looks good!

PR #208, opened 2022-07-21
generate code-signing keys

benadida:


code-signing/generate-keys.sh
```diff
@@ -0,0 +1,64 @@
...
+
+echo -e "\nsigning DB with KEK"
+sign-efi-sig-list -t "$(date --date='1 second' +'%Y-%m-%d %H:%M:%S')" \
+                  -k KEK.key -c KEK.crt db DB.esl DB.auth
+
+chmod 0600 *.key
+
+echo ""
+echo ""
+echo "For use with KeyTool, copy the *.auth and *.esl files to a FAT USB"
```
umbernhard, line 60: Do we need this message? I feel like we should kill it since we're not going to use KeyTool or the firmware interface...

umbernhard (approved):

code-signing/generate-keys.sh
benadida: indeed we don't! I will clean that up.

PR #210, opened 2022-07-25
Add signing capabilities to vx-ui

umbernhard:
I'm not totally sure this is the right way to do it, if there's a better place to put `sign.sh` lmk.


config/sudoers
```diff
@@ -27,5 +27,6 @@ vx-admin ALL=(root:ALL) NOPASSWD: /vx/admin/admin-functions/setup-boot-entry.sh
 vx-ui ALL=(root:ALL) NOPASSWD: /vx/code/config/admin-functions/timedatectl
 vx-ui ALL=(root:ALL) NOPASSWD: /usr/bin/brightnessctl
 vx-ui ALL=(root:ALL) NOPASSWD: /usr/local/bin/tpm2-totp
+vx-ui ALL=(root:ALL) NOPASSWD: /vx/admin/admin-functions/sign.sh
```
benadida, line 30: This is a little odd because `/vx/admin` is meant to be reserved for the `vx-admin` user.

This is also the first time we're giving an admin script to `vx-ui`. It's the right thing, of course, but it's the first time, so we should decide on the right pattern.

A couple of options:
- we copy some of the admin functions to `/vx/ui/admin-functions`
- we create a shared set of admin functions in `/vx/shared/admin-functions`.

The downside of the second option is that the caller would have to know whether the specific executable is in the shared or their own directory, unless we set the path, which seems a little bit brittle.

So my vote would be, let's create `/vx/ui/functions` in `setup-machine`, probably as a new directory rather than a symlink like we have done `/vx/admin/admin-functions`, and copy just `sign.sh` in there for now.

Then, give sudo to `/vx/ui/functions/sign.sh`

benadida (changes requested):

config/sudoers
umbernhard: okay, that should be done. I went with `/vx/ui/ui-functions` for consistency.

benadida (approved):

PR #216, opened 2022-07-27
Fixup TPM code to be more robust to TPM firmware versions

umbernhard:
* Adds support for both SHA1 and SHA256 PCR algorithms
* Doesn't rely on create-load feature, which is only supported on newer devices.

### DIFF
--- a/config/admin-functions/generate-key.sh
+++ b/config/admin-functions/generate-key.sh
@@ -36,7 +36,8 @@ tpm2_evictcontrol -Q -c 0x81000001 || true 2>&1 /dev/null
 tpm2_evictcontrol -c key.ctx 0x81000001
 
 # Delete the local key files
-rm -rf key.pub key.priv
+rm -f key.pub 
+shred -u key.priv
 
 
 rm -f "${VX_CONFIG_ROOT}/key.pub" "${VX_CONFIG_ROOT}/key.sec"
--- a/config/admin-functions/generate-key.sh
+++ b/config/admin-functions/generate-key.sh
@@ -6,8 +6,15 @@ set -euo pipefail
 
 
 set -euo pipefail 
+
+# First detect whether the TPM supports sha1 or sha256
+algo="sha256"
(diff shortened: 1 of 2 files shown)


config/admin-functions/generate-key.sh
```diff
@@ -6,8 +6,15 @@ set -euo pipefail
 
 
 set -euo pipefail 
+
+# First detect whether the TPM supports sha1 or sha256
+algo="sha256"
+if [ $(tpm2 pcrread sha1:0 | wc -l) == 2 ]; then
+    algo="sha1"
```
benadida, line 13: 💩

config/admin-functions/generate-key.sh
```diff
@@ -16,13 +23,17 @@ tpm2_createprimary -C o -c primary.ctx
...
+# Create keys
+tpm2_create -L pcr.policy -u key.pub -r key.priv -C primary.ctx -G ecc 
+tpm2_load -u key.pub -r key.priv -C primary.ctx -c key.ctx
 
 # Save the keys
 tpm2_evictcontrol -Q -c 0x81000001 || true 2>&1 /dev/null
 tpm2_evictcontrol -c key.ctx 0x81000001
 
+# Delete the local key files
+rm -rf key.pub key.priv
```
benadida, line 35: should we `shred` the private key for extra safety?

benadida (approved):

config/admin-functions/generate-key.sh
umbernhard: We can, it is already encrypted. What do you think?

benadida: Yeah I'm not terribly worried, just doing defense in depth you know, in case later we screw up some setting and it doesn't encrypt the private key (why it would do that.. who knows.) Just `shred -u key.priv` is an easy little extra layer of defense.

umbernhard: I'm adding a comment explaining that `key.priv` is encrypted with a key that can't be extracted from the TPM.

PR #218, opened 2022-08-03
Adding Secure Boot status indicator 

umbernhard:
Closes votingworks/vxsuite#1541


config/admin-functions/show-vx-suite-admin-menu.sh
```diff
@@ -39,6 +39,13 @@ while true; do
     echo -e "Lockdown State: \e[31mNot Locked Down\e[0m"
   fi
 
+  if [[ $(mokutil --sb-state | grep "enabled") ]]; then
+    echo -e "Secure Boot State: \e[32mSecure Boot Enabled\e[0m"
```
benadida, line 43: Probably just "enabled" rather than repeating "secure boot" a second time, but otherwise great!

benadida (approved):

PR #222, opened 2022-08-04
Fixes boot entry on Surface Go to use dm-verity

umbernhard:


config/admin-functions/lockdown.sh
```diff
@@ -2,6 +2,14 @@
...
 
 : "${VX_FUNCTIONS_ROOT:="$(dirname "$0")"}"
+: "${VX_CONFIG_ROOT:="/vx/config"}"
+
+# detect Surface Go
+if [ $(cat ${VX_CONFIG_ROOT}/machine-type) == "bsd" ]; then
+    surface=1
+else
+    surface=0
+fi
```
benadida, line 12: this approach is a little bit brittle as it assumes that `bsd` == Surface Go.

Can we instead do the hardware detection like we do in other scripts to determine if the hardware is a Surface Go, independently of what software it's running?

config/admin-functions/lockdown.sh
```diff
@@ -17,25 +25,45 @@ echo "$(cat /vx/admin/config/cmdline)${HASH}" > /tmp/cmdline
...
+    mount /dev/sda /mnt || mount /dev/sda1 /mnt || (echo "Secure boot keys not found; exiting" && sleep 5 && exit);
+
+    sbsign --key=/mnt/DB.key --cert=/mnt/DB.crt --output /boot/efi/EFI/debian/VxLinux-signed.efi /tmp/linux.efi
+
+    # Now install it 
+    bash "${VX_FUNCTIONS_ROOT}/setup-boot-entry.sh"
+else
+    # On a surface we just need to setup the right GRUB entry
+    chmod +w /boot/grub/grub.cfg
+    cp /vx/admin/config/grub.cfg /boot/grub/grub.cfg
```
benadida, line 52: should we make it not writable after this, or maybe just not `chmod` and do a `sudo cp` which ... IIRC preserves the target permissions? I could be wrong on that.

config/admin-functions/lockdown.sh
```diff
@@ -17,25 +25,45 @@ echo "$(cat /vx/admin/config/cmdline)${HASH}" > /tmp/cmdline
...
+    echo "menuentry 'VxLinux' {
+	load_video
+	insmod gzio
+	if [ x\$grub_platform = xxen ]; then insmod xzio; insmod lzopio; fi
+	insmod part_gpt
+	insmod ext2
+       echo 'Loading Linux ${KERNEL_VERSION} ...'
+       linux /vmlinuz-${KERNEL_VERSION} $(cat /tmp/cmdline)
+       echo 'Loading initial ramdisk ...'
+       initrd /initrd.img-${KERNEL_VERSION}
```
benadida, line 63: spacing inconsistency

config/cmdline
```diff
@@ -1 +1 @@
-BOOT_IMAGE=/vmlinuz-5.10.0-11-amd64 root=/dev/mapper/vroot ro quiet vt.handoff=7 fsck.mode=skip verity.hashdev=/dev/mapper/Vx--vg-hashes verity.rootdev=/dev/mapper/Vx--vg-root verity.hash=
+root=/dev/mapper/vroot ro quiet vt.handoff=7 fsck.mode=skip verity.hashdev=/dev/mapper/Vx--vg-hashes verity.rootdev=/dev/mapper/Vx--vg-root verity.hash=
```
benadida, line 1: this removal is purposeful? Just double checking.

benadida (changes requested):
looking good overall, requesting change for the surface go detection approach.

config/admin-functions/lockdown.sh
umbernhard: Yes, I had it configured this way so I could test in a VM. Will fix.

config/admin-functions/lockdown.sh
umbernhard: this whole script is run with sudo. We'd have to chmod -w after the write, which we can do.

config/admin-functions/lockdown.sh
umbernhard: I blame tabs. Will fix.

config/cmdline
umbernhard: I _think_ it's unnecessary and may have been breaking GRUB. I'm testing non-scanner devices atm to verify that it was unnecessary.

umbernhard: Confirmed it's unnecessary

benadida (approved):

PR #227, opened 2022-08-12
Max out precinct scanner volume

jonahkagan:
So we can hear the sound effects!

Not really sure how to test this. Could use some help (or someone to test it out for me)


benadida (approved):
we can only test this in prod builds for now.

PR #230, opened 2022-08-15
fix(audio): add audio packages as part of setup-machine

benadida:


umbernhard (approved):
Only thing that might be worth double checking is if we also need pipewire.

benadida:
I *think* it's not required, but I'll keep it in mind and check!

PR #246, opened 2022-08-25
Revert "feat(terminal): bigger font"

carolinemodic:
Reverts votingworks/vxsuite-complete-system#243

When I create a production image with this change as soon as I boot into tty2 I'm flooded with "Could not get file descriptor referring to the console" errors. Idk if this is something about the build server console or a real problem but I am not figuring it out for m14 so revertting this change for now. 

FYI @benadida


eventualbuddha (approved):

benadida:
uggh I'm sorry about that.

PR #252, opened 2022-09-06
Enable full debug logging in VxDev and route properly

carolinemodic:
This turns on all internal debugging in VxDev automatically, and also fixes the run script such that if you run it with debugging on all of that debugging will automatically pipe to syslog/vx-logs properly instead of just outputting to the console. 

Confirmed that all the other logs were already being routed properly.


umbernhard (approved):
If you wanted a simpler way to do this you could put `export DEBUG=*` in `/etc/profile` and I think it would apply it systemwide

carolinemodic:
Oh good to know. I'm not sure if people will want to sometimes turn off debug logging when running with the script directly so I'll leave it for now but that is good to keep in mind

PR #253, opened 2022-09-07
VxDev minor tweaks

carolinemodic:
Adds the terminal app to the dock in VxDev and updates the default env to have various feature flags turned on.


vxdev/default-env
```diff
@@ -1,3 +1,5 @@
 REACT_APP_VX_DEV=true
-# Uncomment the following line, then run 'Update Code',  to enable write in adjudication in VxDev
-# REACT_APP_VX_ENABLE_WRITE_IN_ADJUDICATION=true
+REACT_APP_VX_ENABLE_WRITE_IN_ADJUDICATION=true
```
umbernhard, line 2: Is this needed for this PR or did it just get included (fine either way, just double checking)

umbernhard (approved):
Looks good!

vxdev/default-env
umbernhard: Oh if I'd read your description you said that it is supposed to be here. My bad

PR #254, opened 2022-09-08
add ability to checkout custom branch on vxdev

carolinemodic:
Adds the option to check vxsuite out to a custom specified branch when building the applications in VxDev

### DIFF
--- a/vxdev/update-code.sh
+++ b/vxdev/update-code.sh
@@ -22,6 +22,9 @@ CHOICES+=('latest')
 echo "2. Latest Stable Release ($LATEST_TAG)"
 CHOICES+=('stable')
 
+echo "3. Custom Branch"
+CHOICES+=('custom')
+
 echo
 read -p "Select Option: " CHOICE_INDEX
 
@@ -30,12 +33,13 @@ then
     echo "You need to select a valid option."
     exit 1
 fi
+
 BRANCH=${CHOICES[$CHOICE_INDEX]}
-sudo apt remove -y nodejs # this will get reinstalled, the version could change based on what branch of the code we are building
+sudo apt remove -y nodejs > /dev/null # this will get reinstalled, the version could change based on what branch of the code we are building
 
 if [[ $BRANCH == 'latest' ]]; then
 	cd vxsuite
-	git checkout main
+ 	git checkout main
 	git pull
 	./script/setup-dev
 	cd ../kiosk-browser
@@ -48,6 +52,21 @@ elif [[ $BRANCH == 'stable' ]]; then
 	git submodule update --init --recursive
 	cd vxsuite
 	./script/setup-dev
+	cd ..
+elif [[ $BRANCH == 'custom' ]]; then
+	read -p "Enter the branch name: " BRANCH_NAME
+	cd vxsuite
+	git checkout main
+	git pull
+	while [ !`git branch -r --list origin/$BRANCH_NAME` ]
+	do
+		read -p "Invalid Branch Name. Try again: " BRANCH_NAME
+	done
+	git checkout $BRANCH_NAME
+	./script/setup-dev
+	cd ../kiosk-browser
+	git checkout main
+	git pull
 	cd ..
 fi
 


vxdev/update-code.sh
```diff
@@ -49,8 +53,26 @@ elif [[ $BRANCH == 'stable' ]]; then
 	cd vxsuite
 	./script/setup-dev
 	cd ..
+elif [[ $BRANCH == 'custom' ]]; then
+	read -p "Enter the branch name: " BRANCH_NAME
+	cd vxsuite
+	git checkout main
+	git pull
+	if [ !`git branch -r --list origin/$BRANCH_NAME` ]
```
eventualbuddha, line 61: Does this work? If you want to redirect the output I suggest doing this:

```suggestion
	if ! git branch -r --list origin/$BRANCH_NAME > /dev/null 2>&1
```

vxdev/update-code.sh
```diff
@@ -49,8 +53,26 @@ elif [[ $BRANCH == 'stable' ]]; then
...
+	fi
+	git checkout $BRANCH_NAME
+	./script/setup-dev
+	cd ../kiosk-browser
+	git checkout main
+	git pull
+	cd ..
 fi
 
+exit 0
```
eventualbuddha, line 75: Intentional?

```suggestion
```

carolinemodic: lol oops, just put this in for testing will remove

vxdev/update-code.sh
```diff
@@ -49,8 +53,26 @@ elif [[ $BRANCH == 'stable' ]]; then
...
+		echo "Invalid Branch Name. Closing in 3 seconds."
+		echo $BRANCH_NAME
+		exit 0
+	fi
+	git checkout $BRANCH_NAME
+	./script/setup-dev
+	cd ../kiosk-browser
+	git checkout main
+	git pull
+	cd ..
```
eventualbuddha, line 72: I generally prefer this style since you don't have to pop back manually:

```suggestion
        (
		cd ../kiosk-browser
		git checkout main
		git pull
	)
```

vxdev/update-code.sh
umbernhard: Yep, this does work fwiw

vxdev/update-code.sh
```diff
@@ -49,6 +53,23 @@ elif [[ $BRANCH == 'stable' ]]; then
...
 	./script/setup-dev
 	cd ..
+elif [[ $BRANCH == 'custom' ]]; then
+	read -p "Enter the branch name: " BRANCH_NAME
+	cd vxsuite
+	git checkout main
+	git pull
+	if [ !`git branch -r --list origin/$BRANCH_NAME` ]
+	then
+		echo "Invalid Branch Name. Closing in 3 seconds."
```
umbernhard, line 63: Would it be better to wrap this in a loop so you don't have to restart the whole script if you make a typo?

vxdev/update-code.sh
```diff
@@ -49,6 +53,23 @@ elif [[ $BRANCH == 'stable' ]]; then
...
+		echo $BRANCH_NAME
+		exit 0
+	fi
+	git checkout $BRANCH_NAME
+	./script/setup-dev
+	cd ../kiosk-browser
+	git checkout main
```
umbernhard, line 70: Are we likely to want to enable kiosk-browser branches too? I suspect not?

umbernhard (changes requested):
I think the branch name entry needs to be a loop, but otherwise this looks good.

vxdev/update-code.sh
carolinemodic: Probably eventually but maybe we'll have merged all the repos by then lol

vxdev/update-code.sh
carolinemodic: updated

vxdev/update-code.sh
```diff
@@ -49,6 +53,21 @@ elif [[ $BRANCH == 'stable' ]]; then
...
 	cd ..
+elif [[ $BRANCH == 'custom' ]]; then
+	read -p "Enter the branch name: " BRANCH_NAME
+	cd vxsuite
+	git checkout main
+	git pull
+	while [ !`git branch -r --list origin/$BRANCH_NAME` ]
+	do
+		read -p "Invalid Branch Name. Try again: " BRANCH_NAME
+	done
```
umbernhard, line 64: awesome, my only nudge might be to make the prompt in the loop a bit clearer about what you're asking for. Instead of "Try again," maybe "Please enter a valid branch name:" or something

umbernhard (approved):

PR #265, opened 2022-11-18
Small VxDev Tweaks

carolinemodic:
Two Small improvements to VxDev 
1. Removes the debug=* flag when running the frontend apps, we added this for full logging but in practice there is SO MUCH logging it slows down the app on VxScan to basically be unusable. The other apps haven't seen as many issues but for consistency I think its better to just remove for now. 
2. Copies the vxdev env file for feature flags to all subrepos. Since the .env file on the subrepos overrides .env.local at the root VxDev users have struggled to set the feature flags properly. There are larger issues with the feature flagging to fix but this will at least make it so that setting things in /vx/config/.env.local will apply as you would expect. 

@amcmanus Don't expect you to review this but cc-ing you for visibility. More context on the env/feature flag conundrum: https://github.com/votingworks/vxsuite/issues/2796

### DIFF
--- a/vxdev/run-vxadmin.desktop
+++ b/vxdev/run-vxadmin.desktop
@@ -2,6 +2,6 @@
 Version=1.0
 Name=VxAdmin
 Icon=runprogram.png
-Exec=gnome-terminal -- /bin/sh -c 'env DEBUG=* env PATH=$PATH:/sbin env ADMIN_WORKSPACE=/vx/data/admin-service /vx/code/vxsuite-complete-system/run.sh election-manager'
+Exec=gnome-terminal -- /bin/sh -c 'env PATH=$PATH:/sbin env ADMIN_WORKSPACE=/vx/data/admin-service /vx/code/vxsuite-complete-system/run.sh election-manager'
 Terminal=false
 Type=Application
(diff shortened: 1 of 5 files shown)


vxdev/run-vxadmin.desktop
```diff
@@ -2,6 +2,6 @@
 Version=1.0
 Name=VxAdmin
 Icon=runprogram.png
-Exec=gnome-terminal -- /bin/sh -c 'env DEBUG=* env PATH=$PATH:/sbin env ADMIN_WORKSPACE=/vx/data/admin-service /vx/code/vxsuite-complete-system/run.sh election-manager'
+Exec=gnome-terminal -- /bin/sh -c 'env PATH=$PATH:/sbin env ADMIN_WORKSPACE=/vx/data/admin-service /vx/code/vxsuite-complete-system/run.sh election-manager'
```
carolinemodic, line 5: sidenote @amcmanus I had so much difficulty figuring out how to get these desktop entry files to launch our run script properly, I could not figure out a way where it would set these variables without having to put them here and also append the sbin to the path (even though it is in the users default path, it would disappear when running through this desktop file). I tried many different ways of calling this with gnome-terminal without that, etc. and this was the ultimate combination of things that seemed to work but I don't understand it and I'm sure there is a more straightforward way to do this.

benadida (approved):
this PR makes me wonder if we need to revisit the way we're doing feature flag / env variable prioritization.... but can be done later.

PR #266, opened 2022-11-30
Support /apps directory structure (for VxScan) 

jonahkagan:
Task: https://github.com/votingworks/vxsuite/issues/2813

Due to the changes in https://github.com/votingworks/vxsuite/pull/2816 which change the vxsuite directory structure for VxScan, scripts in this repo needed to be updated accordingly.

_Review by commits_

Tested by:
- Running `./build.sh vx-scan`
- Running `./run.sh vx-scan`

I know we should probably also test the consumers of those scripts (which also got updated) but I wasn't sure how to test them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ SCAN_WORKSPACE=/tmp ./run.sh bsd
 
 You may replace `/tmp` with any persistent path you like.
 
-### Precinct Scanner (VxScan)
+### VxScan (Precinct Scanner)
 
 This requires some other packages to be installed that, unfortunately, are not
 public. If you have access, go to https://github.com/votingworks/plustekctl and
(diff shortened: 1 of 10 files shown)


carolinemodic (approved):
thank you for also updating the vxdev and setup machine callsites! 

Since you changed the name of what VxDev calls ./build.sh and ./run.sh for I think this will break VxDev's ability to go back in time to a previous tag but... no one uses that right now I'm not sure why we would need to go back in time right now so I think thats ok. You can still manually check out a tag and build/run if you really needed to.

jonahkagan:
@carolinemodic thanks for reviewing! Is it easy for you to check that VxDev works after this change? I don't have it set up on this computer.

carolinemodic:
testing vxdev is always a little weird as you have to comment out the pieces that check out main lol but just tested and it seems to work :)

Issue #268, opened 2022-12-08
make build - broken on a fresh image build


amcmanus:
The addition of build support for the new `apps` structure in vxsuite has introduced a subtle bug only seen when building on a new system.

tl;dr - the easiest fix is to add the packages included in `services/smartcards` `make install` as part of the initial `make deps` step when building vxsuite-complete-system on an image for the first time.

More details:

The latest version of build.sh starts with `apps/vx-scan`. 
Once it hits this command: `BUILD_ROOT="${BUILD_ROOT}/vxsuite" ./script/prod-build`, it invokes a pip dependency that isn't met b/c the `services/smartcards` system packages don't exist since smartcards hasn't been built yet. (Fully admit I don't understand the pip dependency bit in depth, but it's definitely there.)

Note: You'll never encounter this bug on a system that had already installed the packages in `services/smartcards`.

Longer term, I think we should reconsider how we manage system-level packages, along with the `build.sh` script. I think our system's build complexity warrants breaking this script into smaller, more manageable/explicit pieces rather than attempting to meet all needs in a single script. It is a non-trivial exercise to work through the script, especially as our list of systems grow, and the lack of explicit ordering can lead to subtle bugs like this one.

PR #269, opened 2022-12-08
fix dependency issue during build.sh all on a clean system

amcmanus:
Fix for https://github.com/votingworks/vxsuite-complete-system/issues/268

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,16 @@ node:
 	sudo apt install -y curl
 	bash ./setup-node.sh
 
-deps: node
+add-ppa:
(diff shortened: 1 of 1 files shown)


carolinemodic (approved):

eventualbuddha:
These probably weren't being installed manually before. Why is it a problem now? @benadida?

amcmanus:
> 

Before the `apps` directory was included, smartcards got built before anything that might have needed it. (At least that's my current understanding.)

benadida:
> These probably weren't being installed manually before. Why is it a problem now? @benadida?

I'm not entirely sure... maybe the converter-sems was installing it? But since this was one of the more fragile parts of the system, I'm also not terribly surprised it broke and glad we're going to be fixing this up.

eventualbuddha (approved):

PR #272, opened 2023-01-14
New USB Mounting Approach

adghayes:
We want to use `mount` rather than `pmount` for mounting USB drives for two reasons:
- It allows mounting loop devices, which enable us to mock USB drives. Mocking USB drives is important to enable easier development and, down the line, robust integration testing.
- It allows mounting `exFAT`, which we don't need now but may need in the future

Rather than grant users powerful `mount` permissions, we restrict their permissions by granting them access to a limited script which uses `mount` per [@benadida's suggestion](https://github.com/votingworks/kiosk-browser/pull/142#pullrequestreview-1248577513).

Per this change, I opted to move around where our non-admin scripts live. We already have a `sign.sh` script what we can pass into `kiosk-browser` with an argument. This adds more scripts and we may add even more in the future, so instead of passing every single script as an argument to `kiosk-browser`, let's pass in a directory containing all expected scripts. See what this looks like in `kiosk-browser` in https://github.com/votingworks/kiosk-browser/pull/142. I create the directory `app-scripts` and put it at the root. I prefer this to `/config/ui-functions` because semantically it makes more sense (these scripts are not really configuration in the way that most of `/config` is) and, like having `/printing` at the root, it makes it easier to manages paths across our multiple environments. Also, these may be `vx-ui` actions now but could soon be `vx-services` actions. And eventually, we may move `admin-functions` await from `tty2` and then all scripts are `app-scripts`.

### Review Merging
Do not review commit-by-commit, review all together.

Before merging, I will push one more commit that updates the `vxsuite` after https://github.com/votingworks/vxsuite/pull/2892 is merged and the `kiosk-browser` version after https://github.com/votingworks/kiosk-browser/pull/142 is merged.

### Testing
- Confirmed ability to mount and unmount in these environments
  - [x] `kiosk-browser` development, where `kiosk-browser` and `vxsuite-complete-system` are in the same repo
  - [x] Using `kiosk-browser` through `vxsuite-complete-system` in development
  - [x] Production (after `setup-machine.sh`)
  - [ ] Production Locked Down
  - [ ] VxDev
- Confirmed exporting logs to USB via admin menu still works in production
 
I have no reason to suspect that this would break in VxDev or in a Production Locked Down environment because the steps mirror other things we do in those environments.


setup-machine.sh
```diff
@@ -164,21 +164,9 @@ sudo usermod -aG adm vx-services
 # Set up log config
 sudo bash setup-scripts/setup-logging.sh
 
-# Allow mounting of USB
-sudo usermod -aG plugdev vx-ui
-sudo usermod -aG plugdev vx-admin
```
adghayes, line 169: `plugdev` group is specific to `pmount`, no longer necessary.

setup-machine.sh
```diff
@@ -164,21 +164,9 @@ sudo usermod -aG adm vx-services
...
-sudo mkdir -p /media/usb-drive-sdd1
-sudo mkdir -p /media/usb-drive-sde1
-sudo mkdir -p /media/usb-drive-sdf1
-sudo mkdir -p /media/usb-drive-sdg1
-sudo mkdir -p /media/usb-drive-sdh1
-
-sudo chown -R vx-ui:vx-group /media/usb-drive*
+# set up mount point ahead of time because read-only later
+sudo mkdir -p /media/vx/usb-drive
+sudo chown -R vx-ui:vx-group /media/vx
```
adghayes, line 169: If we want to support multiple USB drives in the future, we could do `/media/vx/usb-drive-sda1, /media/vx/usb-drive-sdb1` and so on. I don't see the need to support labels (which are really only the label of the mount point) or multiple USB drives right now, so simplifying until we have requirements.

config/sudoers
```diff
@@ -24,12 +24,16 @@ vx-admin ALL=(root:ALL) NOPASSWD: /vx/admin/admin-functions/set-clock.sh
...
+vx-admin ALL=(root:ALL) NOPASSWD: /usr/local/bin/tpm2-totp
+vx-admin ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/mount.sh
+vx-admin ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/umount.sh
 vx-ui ALL=(root:ALL) NOPASSWD: /vx/code/config/admin-functions/timedatectl
 vx-ui ALL=(root:ALL) NOPASSWD: /usr/bin/brightnessctl
 vx-ui ALL=(root:ALL) NOPASSWD: /usr/local/bin/tpm2-totp
-vx-ui ALL=(root:ALL) NOPASSWD: /vx/ui/ui-functions/sign.sh
+vx-ui ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/sign.sh
+vx-ui ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/mount.sh
+vx-ui ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/umount.sh
```
adghayes, line 35: I think that we could grant `sudo` access either to the original or the symlink and both would work (both already exist in this file) but for simplicity I think granting access to the source is better because then it is the same across users, we could grant perms in one statement if multiple users need the same perms, etc.

config/admin-functions/copy-logs.sh
```diff
@@ -34,8 +33,10 @@ cp -rp /var/log/auth.log* "$DIRECTORY"
 cp -rp /var/log/vx-logs.log* "$DIRECTORY"
 
 # unmount the USB stick to make sure it's all written to disk
-pumount "$DEVICES"
+echo "Saving logs to USB drive..."
+sync $MOUNTPOINT
+sudo /vx/code/app-scripts/umount.sh
```
adghayes, line 38: Fixing a little bug that @mattroe and I have noticed where the logs may not be flushed to the disk.

config/sudoers
```diff
@@ -24,12 +24,16 @@ vx-admin ALL=(root:ALL) NOPASSWD: /vx/admin/admin-functions/set-clock.sh
 vx-admin ALL=(root:ALL) NOPASSWD: /vx/admin/admin-functions/lockdown.sh
 vx-admin ALL=(root:ALL) NOPASSWD: /vx/admin/admin-functions/generate-key.sh
 vx-admin ALL=(root:ALL) NOPASSWD: /vx/admin/admin-functions/setup-boot-entry.sh
+vx-admin ALL=(root:ALL) NOPASSWD: /usr/local/bin/tpm2-totp
+vx-admin ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/mount.sh
+vx-admin ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/umount.sh
```
adghayes, line 29: Necessary for the `copy-logs.sh` script.

amcmanus (approved):
I would suggest renaming `mount.sh` and `umount.sh` to `mount_usb.sh` and `umount.sh` to `umount_usb.sh` since these are very specific to the type of [u]mount operation they perform. Looks good though.

benadida (approved):
lgtm!

PR #273, opened 2023-01-28
move more data off the /var partition in preparation for locking things down more

benadida:
- move home directories for `vx-ui`, `vx-admin`, `vx-services` to the root partition (we might make this the `/home` partition at some point)
- leave just the `.config/kiosk-browser` directory as a symlink to `/var/`
- adjust the permissions accordingly


amcmanus (approved):

benadida:
ahah, this doesn't work yet because the lockdown script, which needs to remount `/` as read-only, and which we run as `vx-admin` which now has its home directory under `/`, gets a mountpoint busy error. I think we probably want to move the home directories to `/home` and actually keep that partition separate from `/` for the sake of being able to do the lockdown.

cc @amcmanus

benadida:
@amcmanus ok got this working all the way through to lockdown, had to move things to `/home`, and in the process tweak what we `rm` at the end of the setup-machine script. Let me know if this looks good to you with these modifications.

benadida:
Hmmm, looking at the next step, which is to `dm-verity` `/home` as well as `/`, I'm not seeing great documentation on how we tell the kernel to verify multiple partitions, and we might need another volume to store the hashes of that second volume... and that all feels unecessary.

So maybe I *do* want to get rid of `/home`, and put everything in `/`, and figure out how to the remounting of `/` as read-only from the lockdown script.

benadida:
OK, these latest changes that put everything under `/` but then have the lockdown script simply cd to `/tmp` before unmounting `/` does the trick! This is great cause we can then remove the separate `/home` partition in a next update, just like @amcmanus was suggesting.

I've confirmed that this works in a VM with lockdown and secure boot, and maintains configuration information across reboots. So I think we're good and actually more locked down with this update!

benadida:
cc @umbernhard for visibility too.

setup-machine.sh
```diff
@@ -130,22 +130,27 @@ sudo mkdir -p /var/vx
...
+sudo mkdir -p /var/vx/ui-kiosk-browser-config
 
 sudo ln -sf /var/vx/data /vx/data
 
 echo "Creating users"
 # create users, no common group, specified uids.
-id -u vx-services &> /dev/null || sudo useradd -u 750 -m -d /var/vx/services vx-services
-id -u vx-ui &> /dev/null || sudo useradd -u 751 -m -d /var/vx/ui -s /bin/bash vx-ui
-id -u vx-admin &> /dev/null || sudo useradd -u 752 -m -d /var/vx/admin -s /bin/bash vx-admin
+id -u vx-services &> /dev/null || sudo useradd -u 750 -m -d /vx/services vx-services
```
umbernhard, line 139: do we need a home directory for these users at all?

PR #276, opened 2023-02-07
Remove references to bas / vxencode

carolinemodic:
BAS aka Ballot-Activation-Service aka VxEncode is an app we have not supported or used in a very long time, we formally killed it in vxsuite here: https://github.com/votingworks/vxsuite/pull/2957

This removes references from complete-system


setup-machine.sh
```diff
@@ -200,7 +196,7 @@ sudo cp config/apparmor.d/usr.sbin.cups-browsed /etc/apparmor.d/
 
 
 # let vx-services scan
-if [ "${CHOICE}" != "vx-mark" ] && [ "${CHOICE}" != "bas" ] 
+if [ "${CHOICE}" != "vx-mark" ]
```
carolinemodic, line 199: @benadida should this also include a skip for VxAdmin? Its enabled scanning functions.

benadida: yeah this was left open for when we thought maybe vxadmin would scan ballots, too, safe to exclude it for now.

amcmanus (approved):

PR #284, opened 2023-04-12
Adjustments for custom america scanner, Rust, and Elo screen

benadida:
@amcmanus your plan for Rust may be quite different, this is what I used for m15a, happy to go a different way if you prefer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,9 @@ node:
 	sudo apt install -y curl
 	bash ./setup-node.sh
 
+rust:
+	curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+
 add-ppa:
 	apt-cache search --names-only "^python3.9$$" | grep python || sudo add-apt-repository -y ppa:deadsnakes/ppa # add deadsnakes only if we need to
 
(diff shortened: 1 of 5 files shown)


setup-machine.sh
```diff
@@ -106,12 +116,11 @@ sudo apt install -y unclutter mingetty brightnessctl alsa-utils pulseaudio pulse
...
-if [[ $DISTRO == "Debian" ]]; then
-	sudo apt install -y rsync cups cryptsetup xserver-xorg-core x11-common xinit sbsigntool
-	sudo chown :lpadmin /sbin/lpinfo
-	echo "export PATH=$PATH:/sbin" | sudo tee -a /etc/bash.bashrc
-fi
+sudo apt install -y xorg xserver-xorg-core xserver-xorg-video-all xserver-xorg-input-all xinput x11-common xinit
+sudo apt install -y rsync cups cryptsetup sbsigntool
+
+sudo chown :lpadmin /sbin/lpinfo
+echo "export PATH=$PATH:/sbin" | sudo tee -a /etc/bash.bashrc
```
benadida, line 123: used the opportunity to not try to be Ubuntu compatible anymore because why, we're not trying elsewhere.

Makefile
```diff
@@ -15,7 +18,7 @@ install-python: add-ppa
 install-smartcard:
 	sudo apt install -y libusb-1.0-0-dev libpcsclite-dev pcscd pcsc-tools swig
 
-deps: node install-python install-smartcard
+deps: node rust install-python install-smartcard
```
carolinemodic, line 21: since we were going to try to not use make deps at all I don't think this conflicts with the build-system approach so I'm happy to approve this for now to unblock you @benadida

carolinemodic (approved):

amcmanus (approved):

PR #292, opened 2023-04-13
play startup chime in background 

benadida:
so as not to delay the screen orientation adjustment


arsalansufi (approved):

arsalansufi (approved):
Follow-up LGTM!

PR #301, opened 2023-05-04
Add sudoers config for libs/usb-drive scripts

jonahkagan:
Give `vx-ui` user permission to execute the new mount/unmount scripts introduced in https://github.com/votingworks/vxsuite/pull/3378


amcmanus (approved):

arsalansufi:
Going to merge this as I'm about to create alpha images for the SLI test (and am 99% sure we need this change for USB interactions to work in VxScan)

PR #302, opened 2023-05-11
Use TPM key to create machine cert and for card programming

arsalansufi:
_Easiest reviewed by commit_

## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/3055

The last step in using the TPM key to create machine certs and for subsequent card programming (as opposed to a private key file)! Ties together the foundation laid in https://github.com/votingworks/vxsuite/pull/3433 and https://github.com/votingworks/vxsuite/pull/3436.

## Testing

- [x] Created an image and manually tested configuration scripts and card programming on real hardware


setup-scripts/setup-tpm2-tools.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
+
 set -euo pipefail 
 
-sudo apt -y install tpm2-tools qrencode
+# Where libengine-tpm2-tss-openssl lives
+sudo sh -c 'echo "\ndeb http://http.us.debian.org/debian bullseye-backports main" >> /etc/apt/sources.list'
```
umbernhard, line 6: We might want to be more careful about throwing in backports like this. I would recommend (for now) adding it but pinning its priority to really low, and then using `apt install -t bullseye-backports libengine-tpm2-tss-openssl` to make sure that *only* that package is retrieved from backports. 

The other potential complication here is that if we bump versions to Debian 12, the backport will break (tho it will not be necessary any more, since the stuff in bullseye-backports is stuff that's in 12 that is ported back to 11).

config/sudoers-for-dev
```diff
@@ -30,6 +30,8 @@ vx-admin ALL=(root:ALL) NOPASSWD: /vx/admin/admin-functions/program-system-admin
 vx-admin ALL=(root:ALL) NOPASSWD: /usr/local/bin/tpm2-totp
 vx-admin ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/mount-usb.sh
 vx-admin ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/unmount-usb.sh
+vx-admin ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/create-cert
+vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/create-cert
```
umbernhard, line 34: Do both of these users need to be able to run these scripts?

config/admin-functions/create-machine-cert.sh
```diff
@@ -79,38 +79,20 @@ function wait_for_usb_and_mount_once_present() {
...
-    openssl req -new \
-        -config "${VX_METADATA_ROOT}/vxsuite/libs/auth/certs/openssl.cnf" \
-        -key "${MACHINE_PRIVATE_KEY_PATH}" \
-        -passin "pass:${machine_private_key_password}" \
-        -subj "${machine_cert_subject}" \
-        -out "${cert_signing_request_path}"
+    local machine_jurisdiction="${1:-}"
+    if [[ -n "${machine_jurisdiction}" ]]; then
+        # Use a subshell to ensure the cd doesn't have an effect beyond this command
+        (cd "${VX_METADATA_ROOT}/vxsuite/libs/auth" &&
```
umbernhard, line 86: you could also use `pushd` and `popd` if you wanted, but I think a subshell is fine?

umbernhard (approved):
I think this looks good!

config/sudoers-for-dev
arsalansufi: Yeah, thanks for checking! They do - the explanation is tucked away in the commit message:

> \- vx-admin user needs this for sys admin card programming in configuration wizard
> \- vx-services user needs this for card programming in general

config/admin-functions/create-machine-cert.sh
arsalansufi: Nice! I like `pushd` and `popd` for clarity without an explanatory comment: https://github.com/votingworks/vxsuite-complete-system/pull/302/commits/43cff6f94ba3b2ee102c04869c5c8bc0b1de1f0c

setup-scripts/setup-tpm2-tools.sh
arsalansufi: Good call! Took the approach that you suggested for now: https://github.com/votingworks/vxsuite-complete-system/pull/302/commits/0dc4f14f01ebaed0b62874176c35c88202d6ab25

PR #303, opened 2023-05-20
move to mimic3 speech synthesis, with caching

benadida:


config/ui-functions/mimic3-cache
```diff
@@ -0,0 +1,26 @@
...
+hash=`echo "${input}" | sha1sum | cut -d' ' -f1`
+filename="${CACHE_DIR}/${hash}.wav"
+
+# if no file or the file is empty (can happen on crash in previous attempts to generate this file), generate it
+if ! [ -s "${filename}" ]; then
+    echo "${input}" | /usr/bin/mimic3 --remote --voice en_US/m-ailabs_low --stdout > ${filename}
+fi
+
+# just return the filename of the cached file
+echo `realpath ${filename}`
```
eventualbuddha, line 26:
```suggestion
realpath ${filename}
```

config/ui-functions/mimic3-cache
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+CACHE_DIR=${CACHE_DIR:-/vx/ui/mimic-cache}
```
eventualbuddha, line 3: While using speech-dispatcher and caching seems like the smaller change to make, I wonder whether you tried running mimic3 as an HTTP server that can be queried. If so, is it fast enough for real-time? I remember you saying the startup time was a big part of the problem.

config/ui-functions/mimic3-cache
```diff
@@ -0,0 +1,26 @@
...
+# trim leading and trailing spaces
+input=$(echo "$input" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+
+# have we cached this?
+hash=`echo "${input}" | sha1sum | cut -d' ' -f1`
+filename="${CACHE_DIR}/${hash}.wav"
+
+# if no file or the file is empty (can happen on crash in previous attempts to generate this file), generate it
+if ! [ -s "${filename}" ]; then
+    echo "${input}" | /usr/bin/mimic3 --remote --voice en_US/m-ailabs_low --stdout > ${filename}
```
eventualbuddha, line 22: These never get cleaned up, right? Seems like a LRU cache would be in order, or perhaps a cache that gets cleared on election unconfigure.

config/vx-mimic3.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Mimic3 Speech Synthesis
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/mimic3-server --voice en_US/m-ailabs_low
```
eventualbuddha, line 6: Oh, maybe you are using the server? As best I can tell from a quick read, `/usr/bin/mimic3` just calls into [`mimic3_tts`](https://github.com/MycroftAI/mimic3/blob/master/mimic3_tts/__main__.py#L348) and does not use the server. Is this actually used?

config/mimic3-generic.conf
```diff
@@ -0,0 +1,2 @@
+GenericExecuteSynth "/vx/ui/ui-functions/spd-mimic3-backend -w '$DATA' | $PLAY_COMMAND"
```
eventualbuddha, line 1: I'm surprised this single-quoted argument gets expanded. Perhaps `speech-dispatcher` is doing the shell expansion rather than the shell itself?

I worry a little bit about arbitrary text being put into these bash variables…

benadida: yes, I think `speech-dispatcher` is doing this.

You're totally right about the concern here... we could use some input sanitization here as a layer of defense.

config/ui-functions/mimic3-cache
benadida: yep, that is also happening, as you see below. There are three things making things slow:

- loading the voice model
- synthesizing an utterance
- starting up the python mimic3 client

The first is addressed by running the server – I've confirmed that my setup doesn't work if I turn off the server

The second is address by pre-warming the cache

The third is addressed by putting the cache in *front* of mimic3.

config/ui-functions/mimic3-cache
benadida: indeed, one of my next quick-and-dirty cleanup ideas that I want to add is, in mimic3-cache, simply remove any files that are older than 2 months.

benadida: Also, not the `--remote`, that's the part that makes `mimic3` contact the http server instead of loading everything itself.

config/vx-mimic3.service
benadida: I'm fairly certain that the `--remote` option is what makes the `mimic3` client contact the server.

config/ui-functions/mimic3-cache
```diff
@@ -0,0 +1,29 @@
...
+
+switch=$1
+input=$2
+
+if [[ $switch != "-w" ]]; then
+    exit 0
+fi
+
+# clear files older than 2 months from the cache
+find ${CACHE_DIR} -type f -name "*.wav" -mtime +60 -delete
```
benadida, line 14: @eventualbuddha cache clearing fix.

carolinemodic (approved):

PR #308, opened 2023-05-28
bump vxsuite for rc3

benadida:


benadida:
merging cause it's just a bump.

PR #310, opened 2023-06-06
Fix build script

carolinemodic:
With the design app only having a frontend repo the build script breaks when building all apps because there is no backend directory, this should fix that.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -56,6 +56,10 @@ usage() {
 # Function that builds a single app
 build() {
   local APP="$1"
+  # temporary hack to fix the fact that the design app is not meant for production
+  if [ "${APP}" = "design" ]; then
+    return
+  fi
   echo "🔨Building ${APP}"
   export BUILD_ROOT="${DIR}/build/${APP}"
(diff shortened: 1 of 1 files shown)


benadida (approved):

carolinemodic:
had to take a hackier approach post-review after realizing other issues, cc @kofi-q I think we should move design out of the "apps" sub directory and keep that for apps we want built for production purposes but this will avoid the issue for now.

arsalansufi:
To avoid future confusion, `apps/design` actually holds @jonahkagan's ballot design UI and not any of @kofi-q's design work (which is mostly contained to `libs/ui`). Very reasonable interpretation of `apps/design` though, @carolinemodic, and maybe a sign that we need a more descriptive name!

carolinemodic:
ohhh makes sense haha I didn't actually look that closely. @jonahkagan when you return we should figure out the right solution here if you app should be supported as a first class citizen in complete system or be in a different sub dir

jonahkagan:
Sounds good, we can discuss

PR #313, opened 2023-06-09
tweak for elo13

benadida:
I'll test this on real hardware tomorrow.

PR #314, opened 2023-06-09
fix: permissions for USB mounting and sound on Asus

benadida:


config/sudoers
```diff
@@ -32,6 +32,8 @@ vx-admin ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/unmount-usb.sh
 vx-admin ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/intermediate-scripts/create-cert
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/intermediate-scripts/create-cert
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/intermediate-scripts/sign-message
+vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/usb-drive/src/mount.sh
+vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/usb-drive/src/unmount.sh
```
arsalansufi, line 36: I _think_ we can also get rid of the vx-ui permissions for these scripts. Worth testing to be sure!

Also a reminder to update sudoers-for-dev as well

benadida: oh good call on sudoers-dev.

I was thinking of leaving it for vx-ui... are we sure every component is using vx-services for this now?

arsalansufi: I'm 90% sure we don't need it for vx-ui since kiosk-browser USB mounting/unmounting, though still in use, uses /vx/code/app-scripts/mount-usb.sh etc. (and not /vx/code/vxsuite/libs/usb-drive/src/mount.sh etc.). But agree that it's fine for now to just keep it for both. We can clean up as we continue to consolidate USB logic!

config/xinitrc
```diff
@@ -12,19 +12,20 @@ export VX_METADATA_ROOT="/vx/code"
 
 # Elo screen, adjust orientation and corresponding touch coordinates
 # xinput is not ready to be used until X has fully started, so we have to delay
-ELO="Elo Touch ET1502L"
-(sleep 4 && if xinput list | grep "$ELO" > /dev/null; then
+ELO="Elo Touch"
+(sleep 4 && if xinput list | grep -i "$ELO" > /dev/null; then
```
arsalansufi, line 16: Added this `-i` (for case insensitivity) since I noticed it in https://github.com/votingworks/vxsuite-complete-system/pull/313. May not be necessary but am playing it safe

arsalansufi (approved):

PR #322, opened 2023-07-04
making home read-only and making X compatible with it

benadida:
This makes all home directories read-only as follows:
- remove the separate `/home` partition, it's now part of `/` – update the pre-seed accordingly
- tell Xwindows to use `/tmp/.Xauthority` as its lock file
- give a writable space for `/vx/ui/.local` so X can log there to its heart's content
- make sure all home directories are now at `/vx/...` rather than `/var/vx/...`

While we're at it:
- remove some permissions so it's harder for `vx-ui` to run a node executable. (This needs more work, but it's a start.)

The changes above make it more difficult to lock down, because the lockdown procedure requires remounting `/` as read-only, and that can't happen when we're using files in `/`, which we now are cause that includes the home directories.

So lockdown is now in two phases:
- from `tty2` menu, choose lock down on next reboot, which sets a flag and updates `fstab` to remount `/` as read-only
- on next boot, menu automatically comes up, choose option to lock down. That performs the actual lockdown, and since `/` is already mounted read-only, all is well.

### DIFF
--- a/config/ui_bash_profile
+++ b/config/ui_bash_profile
@@ -1,7 +1,7 @@
-export PATH=/vx/code/config/admin-functions:${PATH}
+export PATH=/vx/code/config/admin-functions:/sbin:${PATH}
 
 if [[ $(tty) = /dev/tty1 ]]; then
-   exec startx
+   # if startx crashes or quits for any reason, log out
+   XAUTHORITY=/tmp/.Xauthority startx || logout
 fi
 
-export PATH=$PATH:/sbin
--- a/config/admin-functions/generate-key.sh
+++ b/config/admin-functions/generate-key.sh
@@ -10,39 +10,45 @@ if [ "$(tpm2 pcrread sha256:0 | wc -l)" == "2" ]; then
     algo="sha256"
 fi
 
-tpm2_startauthsession -S session.ctx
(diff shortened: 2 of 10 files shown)


config/ui_bash_profile
```diff
@@ -1,7 +1,7 @@
 export PATH=/vx/code/config/admin-functions:${PATH}
 
 if [[ $(tty) = /dev/tty1 ]]; then
-   exec startx
+   XAUTHORITY=/tmp/.Xauthority startx
```
eventualbuddha, line 4: This no longer replaces the current process with `startx`, so if `startx` exits it'll continue the script. Is that intended?

benadida: Intended yes, but not as a final end-state, I'm playing with a few other things.

Are we certain that with `exec` it will cause a logout if you ctrl-C while it boots?

eventualbuddha: I'm not sure how `ui_bash_profile` is run, so I'm not sure about its behavior with regard to logouts. I'm just saying that this will print `foo`:

```sh
run_something_that_exits
echo foo
```

and this will not:

```sh
exec run_something_that_exits
echo foo
```

arsalansufi (approved):

PR #327, opened 2023-07-10
setup, scripts, and permissions to switch apps from vxadmin to vxcentralscan and back

benadida:
Currently, in production, we only copy the run scripts and services for the component we've configured. That's probably the right thing to do.

For this prototype, we copy them all, so we can easily switch from vxadmin to vxcentralscan.


run-scripts/run-kiosk-browser-forever-and-log.sh
```diff
@@ -28,6 +28,7 @@ if [ "${VX_MACHINE_TYPE}" = "scan" ]; then
 fi
 
 while true; do
+    sleep 2
```
carolinemodic, line 31: lol

setup-machine.sh
```diff
@@ -342,9 +356,22 @@ sudo rm -rf /lib/modules/*/kernel/drivers/net/*
...
+    if [ -f "config/${service}.service" ]; then    
+	sudo cp config/${service}.service /etc/systemd/system/
+	sudo chmod 644 /etc/systemd/system/${service}.service
+    fi
+done
+
+# watch and switch service
+sudo cp config/watch-and-switch.service /etc/systemd/system/
+sudo chmod 644 /etc/systemd/system/watch-and-switch.service
+sudo systemctl enable watch-and-switch.service
```
carolinemodic, line 371: so this will be running on scan too? Any potential issues with that?

carolinemodic (approved):

run-scripts/run-kiosk-browser-forever-and-log.sh
benadida: in fact I am going to move this to after....but yeah :)

setup-machine.sh
benadida: that's the one out-there risk I was mentioning... but I could just not do this on scan. Let me fix that.

carolinemodic (approved):

PR #331, opened 2023-07-14
Dont preserve permissions when copying logs

carolinemodic:
Will see if this fixes the tty2 export


adghayes (approved):

benadida (approved):
confirms, works in prod image

PR #352, opened 2023-12-05
fix flicker on raptor lake cards

amcmanus:
Under X, modesetting isn't able to load correctly due to firmware not being loaded properly. That results in initial screen flickering when kiosk-browser starts/restarts. This setting corrects that and does not interfere on systems with a different intel card.


carolinemodic (approved):
hooray!

PR #354, opened 2023-12-07
Change approach to kernel module options and potential signing requirements

amcmanus:
To eliminate screen flicker on the Raptor Lake card, we need to set `i915.enable_guc=2` to ensure all firmware loads. Prior to secure boot, we passed the option via the grub cmdline. The initial secure boot build brought back the screen flicker. Rather than also adding `i915.enable_guc=2` to the secure boot boot entry (feels very fragile, easy to forget on future module options, have to edit in two places, etc...), switched to using modprobe to configure any necessary options. This will be easier to maintain and understand. Since we're using modprobe, the option was removed from the non-secure boot grub config. 

During the troubleshooting phase, we also thought we might need to sign the `i915` module. That turned out to be a red herring since it's already secure boot signed via the Debian key. I've left in code to sign kernel modules in the future though, in case we ever end up needing one that is unsigned. 

If we find ourselves having to sign modules frequently, it will likely be worth a small refactor so module names are not hardcoded in the lockdown script, but this is sufficient for now.

### DIFF
--- a/config/admin-functions/lockdown.sh
+++ b/config/admin-functions/lockdown.sh
@@ -26,7 +26,12 @@ if [[ $answer == 'n' || $answer == 'N' ]]; then
     exit
 fi
 
-modules_to_sign="i915"
+# We don't need to sign i915 since it is signed by Debian's Secure Boot key
+# and we have access to that under Secure Boot. 
+# However, if we do ever need to use an unsigned module, the below code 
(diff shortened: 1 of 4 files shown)


carolinemodic:
going to defer to @benadida here

benadida (approved):

PR #355, opened 2023-12-11
Add image signature option to admin menu

amcmanus:
Once a trusted build has been locked down, a hash of the read-only root partition needs to be capable of being generated. This PR adds the functionality, along with necessary permissions and menus.


config/admin-functions/hash-signature.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+echo "Retrieving image signature..."
```
benadida, line 3: Should we say that this might take some time?

benadida (approved):

config/admin-functions/hash-signature.sh
amcmanus: Good call.

PR #359, opened 2023-12-17
fix(configure): we don't need to show the public key or force it to be recorded

benadida:


benadida:
> You're sure the signify script isn't called from anywhere?

Quite confident, but I put @arsalansufi on this to double check.

Signify is what we used for signing before we moved to openssl-based signatures.

carolinemodic (approved):

benadida:
for posterity, removed the removal of signify, to reduce any risk from changes.

PR #362, opened 2023-12-18
fix(apparmor): remove profile for cups-browsed which we don't install on prod machines

benadida:
The AppArmor profile causes, on current prod builds, an error message on boot that looks scary but is actually a no-op, since we don't have `cups-browsed` as a package anymore. This PR just removes that no-longer-used profile.


carolinemodic:
@benadida did we test this? Do we want to try to land or abandon?

benadida:
> @benadida did we test this? Do we want to try to land or abandon?

This one is very safe to merge.

amcmanus (approved):
I thought I had already approved this...

PR #365, opened 2024-02-10
disable kiosk printing for all apps but mark

adghayes:
it will default to on for development. I haven't seen any issues with it in dev while the backend service is also managing printing, but wouldn't risk it in prod

### DIFF
--- a/run-scripts/run-kiosk-browser.sh
+++ b/run-scripts/run-kiosk-browser.sh
@@ -3,16 +3,24 @@
 set -euo pipefail
 
 URL=${1:-http://localhost:3000}
-: "${VX_CONFIG_ROOT:="./config"}"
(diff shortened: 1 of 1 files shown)


run-scripts/run-kiosk-browser.sh
```diff
@@ -3,16 +3,24 @@
 set -euo pipefail
 
 URL=${1:-http://localhost:3000}
-: "${VX_CONFIG_ROOT:="./config"}"
-: "${VX_METADATA_ROOT:="./"}"
-
-OS=$(lsb_release -cs)
```
adghayes, line 9: These look like they don't need to be here.

benadida:
I love this progress.

eventualbuddha (approved):

PR #366, opened 2024-02-15
Add mark scan setup machine

carolinemodic:
Adds the bare minimum to allow us to make MarkScan images with setup-machine and then go through the certifier dance for them. I imagine there will be other things we need to add to the setup machine script for setting up the accessible controller and such

### DIFF
--- a/config/admin-functions/mock-vx-certifier.sh
+++ b/config/admin-functions/mock-vx-certifier.sh
@@ -20,6 +20,7 @@ if [[
     "${VX_MACHINE_TYPE}" != "admin" &&
     "${VX_MACHINE_TYPE}" != "central-scan" &&
     "${VX_MACHINE_TYPE}" != "mark" &&
+    "${VX_MACHINE_TYPE}" != "mark-scan" &&
     "${VX_MACHINE_TYPE}" != "scan"
 ]]; then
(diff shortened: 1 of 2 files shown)


config/admin-functions/mock-vx-certifier.sh
```diff
@@ -20,6 +20,7 @@ if [[
     "${VX_MACHINE_TYPE}" != "admin" &&
     "${VX_MACHINE_TYPE}" != "central-scan" &&
     "${VX_MACHINE_TYPE}" != "mark" &&
+    "${VX_MACHINE_TYPE}" != "mark-scan" &&
```
arsalansufi, line 23: There are a few cert-related tweaks we need to make to vxsuite, too, e.g., here:

https://github.com/votingworks/vxsuite/blob/95c277b101c74aa1e1ee5490bb77c7f5e7cdaae4/libs/auth/src/certs.ts#L140-L149

I can open a PR for those!

arsalansufi (approved):

kshen0 (approved):

config/admin-functions/mock-vx-certifier.sh
arsalansufi: https://github.com/votingworks/vxsuite/pull/4604

PR #367, opened 2024-04-02
Add support for mark-scan daemons to the trusted build process

amcmanus:
This PR includes mark-scan daemons (accessible-controller and pat-device-input) in the build process, and meets trusted build requirements. 

It also adds the `vx-ui` and `vx-services` users to the `audio` group to support `amixer` access.

Finally, it updates the vxsuite submodule to latest main.

In the current state, both daemons fail to run due to unrelated issues. Since those daemons are still in active development, it's fine if we hold off on merging this to main and update the vxsuite submodule one more time once daemons are ready.

### DIFF
--- a/config/xinitrc
+++ b/config/xinitrc
@@ -24,8 +24,8 @@ PENMOUNT="PENMOUNT PM2501"
      pactl unload-module module-suspend-on-idle
 
      # we are using landscape now, the following is for portrait
-     # xrandr -o left
-     # xinput set-prop "$PENMOUNT" 'Coordinate Transformation Matrix' 0 -1 1 1 0 0 0 0 1
+     xrandr -o left
+     xinput set-prop "$PENMOUNT" 'Coordinate Transformation Matrix' 0 -1 1 1 0 0 0 0 1
 
      # we need a startup chime to activate the channel
      # delay because it appears the xrandr and xinput commands above turn off sound for a bit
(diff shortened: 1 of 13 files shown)


setup-machine.sh
```diff
@@ -137,6 +137,10 @@ sudo usermod -aG vx-group vx-services
 
 sudo usermod -aG video vx-ui
 
+# mark-scan requires access to the audio group
```
kshen0, line 140: @kofi-q does `mark` also require access? It doesn't seem like this config is specific to a single app so probably not an issue, but wanted to check.

kshen0 (approved):

setup-machine.sh
kofi-q: Sorry for the delay! Yeah, audio access is also needed for `mark`, so this would apply to that as well

config/xinitrc
```diff
@@ -10,10 +10,10 @@ export VX_METADATA_ROOT="/vx/code"
 # this appears to fix the issue where, sometimes, the Elo touchscreen doesn't respond
 (sleep 6 && bash /vx/code/run-kiosk-browser-forever-and-log.sh) &
 
-# Elo screen, adjust orientation and corresponding touch coordinates
```
kshen0, line 13: We don't currently use any Elo screens in portrait mode

kshen0 (approved):

PR #369, opened 2024-05-07
add basic restart behavior to markscan daemons

amcmanus:
Adds some basic automatic restart behavior to the vxmarkscan daemons to eliminate potential race conditions during boot.


kshen0 (approved):
Tested this on a prod-imaged machine 👍

PR #370, opened 2024-05-09
Updates for signed hash validation

amcmanus:
This PR adds a noninteractive option to the `hash-signature.sh` script used to generate a hash of the read-only root partition. `sudo` privileges required to run this script are also granted to the `vx-services` user. 

To run as the `vx-services` user in noninteractive mode:
```
sudo /vx/admin/admin-functions/hash-signature.sh noninteractive
```


config/sudoers
```diff
@@ -35,6 +35,7 @@ vx-admin ALL=(root:ALL) NOPASSWD: /usr/sbin/reboot
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/intermediate-scripts/create-cert
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/intermediate-scripts/sign-message
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/usb-drive/scripts/*
+vx-services ALL=(root:ALL) NOPASSWD: /vx/admin/admin-functions/hash-signature.sh
```
arsalansufi, line 38: I think we'll want to update `config/sudoers-for-dev`, too

arsalansufi (approved):
One quick comment then LGTM!

config/sudoers
amcmanus: Good catch, almost always forget about this file.

PR #371, opened 2024-05-16
remove a lot of kiosk-browser functionality

adghayes:
removes printing from `kiosk-browser`, among other things, which means some config pieces in `complete-system` can be removed too. updates pins to `kiosk-browser` and `vxsuite`


setup-machine.sh
```diff
@@ -156,11 +156,9 @@ sudo bash setup-scripts/setup-logging.sh
 sudo mkdir -p /media/vx/usb-drive
 sudo chown -R vx-ui:vx-group /media/vx
 
-# let vx-ui and vx-services manage printers
-sudo usermod -aG lpadmin vx-ui
+# let vx-services manage printers
 sudo usermod -aG lpadmin vx-services
```
adghayes, line 160: Frontend doesn't need these permissions anymore!

eventualbuddha (approved):

PR #380, opened 2024-07-11
Minor updates to get vxdev closer to prod settings

amcmanus:
This PR adds a few of the more recent configs made to prod that weren't synced over to the vxdev setup. We should see if there's a way to keep these in sync more automatically rather than requiring always remembering to make updates in both locations.


carolinemodic (approved):
I feel like there should be some script that both VxDev and Prod Images call where anything that should be shared should go and then this script and the setup-machine script should just do things that are specific to either vxdev or prod images

PR #384, opened 2024-07-17
Fix libs/backend system calls in prod

arsalansufi:
See https://github.com/votingworks/vxsuite/pull/5070 for more context. Opened https://github.com/votingworks/vxsuite/issues/5068 for a future time.

### DIFF
--- a/config/sudoers
+++ b/config/sudoers
@@ -36,6 +36,7 @@ vx-admin ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/intermediate-sc
 vx-admin ALL=(root:ALL) NOPASSWD: /usr/sbin/reboot
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/intermediate-scripts/create-cert
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/intermediate-scripts/sign-message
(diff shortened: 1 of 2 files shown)


adghayes:
@arsalansufi it looks to me like we're no longer using our custom `timedatectl` script, so it could be removed along with its special sudo privileges.

What appears to be the default `timedatectl` is called in two places: in `setup-machine.sh` before lockdown and in `set-clock.sh` during configuration. In that second case, based on [the reasoning here](https://github.com/votingworks/vxsuite/pull/5070/commits/a1c86deb9260762d7669e83b3f28c0db895e9b27#r1681859142),  shouldn't it fail in a locked down environment?

What am I missing - are we somehow sneakily calling our own `timedatectl` in `set-clock.sh`, or is the default `timedatectl` working despite the read-only symlink indirection?

arsalansufi:
@adghayes Great question that forced me to dig a bit deeper

> are we somehow sneakily calling our own `timedatectl` in `set-clock.sh`?

This is it! In the `.bash_profile` for the `vx-admin` user, we update the `PATH` to include the directory containing the custom `timedatectl` function. And because we include that directory first, binaries in that directory will take precedence over system binaries.

https://github.com/votingworks/vxsuite-complete-system/blob/14878282277ee2782b3379e10b39c067ad84d235/config/admin_bash_profile#L1

On that note, we should keep the custom `timedatectl` around for now but can likely remove the `vx-ui` `sudoers` entry for it. I do want to clean things up holistically and am tracking that [here](https://github.com/votingworks/vxsuite/issues/5068).

amcmanus (approved):

PR #388, opened 2024-07-25
Fix vxscan flicker

amcmanus:
This PR implements a fix for initial screen flickering after initial boot or resetting kiosk-browser. As noted in `setup-machine.sh`, this is not an ideal long term fix since we are likely to have `mark` and/or `mark-scan` systems with an intel card that experiences this flicker. For now though, putting this configuration on our other machine types resolves the problem for the foreseeable future. 

The xorg config changes the acceleration method from the default of `sna` (the newer backend) to `uxa` (the previous, more backwards compatible backend). We've run into other flickering on certain intel cards that were fixed with setting options on the i915 module, but those were insufficient on our latest VxScan / VxComputer hardware.

### DIFF
--- a/config/10-intel-xorg.conf
+++ b/config/10-intel-xorg.conf
@@ -1,0 +1,5 @@
+Section "Device"
+	Identifier "Intel Graphics"
+	Driver "modesetting"
+	Option "AccelMethod" "uxa"
+EndSection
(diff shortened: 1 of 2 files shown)


carolinemodic (approved):

arsalansufi (approved):
LGTM! Really glad you were able to figure this out so quickly

PR #389, opened 2024-07-26
add run script and config for fai-100 daemon

kshen0:
Adds service config and run script for the FAI-100 daemon on the BMD 150. The FAI 100 replaces both `controllerd` and `patinputd` on the BMD 155 model.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -55,7 +55,7 @@ if [[ " ${ALL_APPS[@]} " =~ " ${APP} " ]]; then
   # reload their daemon configs and then issue restart commands
   if [[ "${APP}" == "mark-scan" ]]; then
     sudo systemctl daemon-reload
-    for vx_daemon in controller pat
+    for vx_daemon in controller pat fai-100
     do
       sudo systemctl restart mark-scan-${vx_daemon}-daemon.service
     done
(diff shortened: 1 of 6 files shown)


run.sh
```diff
@@ -55,7 +55,7 @@ if [[ " ${ALL_APPS[@]} " =~ " ${APP} " ]]; then
   # reload their daemon configs and then issue restart commands
   if [[ "${APP}" == "mark-scan" ]]; then
     sudo systemctl daemon-reload
-    for vx_daemon in controller pat
+    for vx_daemon in controller pat fai-100
```
kshen0, line 58: @amcmanus any recommendation for bifurcating this? We don't need `controller` and `pat` to run for BMD 150 models, and we don't need `fai-100` for BMD 155 models.

The app uses an env var `MARK_SCAN_USE_BMD_150` to determine what hardware it's running on. Is it possible to set that env var during the image creation process, prior to `run.sh` being called, and check it here?

amcmanus: @kshen0, I've added checks where necessary that depend on the `REACT_APP_VX_MARK_SCAN_USE_BMD_150` var in the vxsuite `.env` file.

kshen0 (commented):
Logic to handle `REACT_APP_VX_MARK_SCAN_USE_BMD_150` LGTM

amcmanus (approved):

PR #392, opened 2024-08-21
add rsyslog configs to forward more logs to vx-logs per vvsg2 requirements

amcmanus:
This PR addresses several logging requirements needed for VVSG2 compliance. sudo actions, password changes, and dm-verity boot information are all now forwarded to the `vx-logs.log` log in json format.


carolinemodic (approved):
Yay! Not urgent as it won't break anything rn, but can you also add any new "event-id" values to this file (then run pnpm build to autogen others) that will add it to the list of logs we have autogenerated docs for and we will probably want it for the cdf conversion as well. 

https://github.com/votingworks/vxsuite/blob/main/libs/logging/log_event_details.toml

PR #393, opened 2024-09-05
Automatically resize /var to max available disk space

amcmanus:
This PR automates the expansion of the partition `/var` resides on to the maximum amount of unallocated disk space on a system. This enables the use of smaller base images (we now use a 27gb base image for all builds, rather than 110gb for everything except VSAP). As a result, creating, cloning, encrypting, and flashing images is significantly faster. It also means our systems can take full advantage of available disk space for logs, CVRs, ballot images, etc... rather than being artificially limited to a 110gb partition structure. 

Of note:

* This works on VxDev and regular images. VxDev implements the resize as a `.profile` hook, and regular images include it as part of the initial boot configuration process. 
* A flag file is created to prevent repeated executions, but the script can be run repeatedly without impacting the system if the flag is removed. 
* Two assumptions about our partition scheme are made: we use a `Vx` naming convention, and the partition to expand is `3`. If either of these assumptions were to change, we will have to update the resize process. 
* By default, LVM uses `/etc/lvm` to store some minimal information. It is not necessary for LVM functionality since volume specific metadata is stored on the physical volumes, but it does have to be accessible when performing certain LVM operations. On a locked-down system, we cannot write to `/etc/lvm`, so we override this default and use `/home/.lvm`. I've chosen to make that the default whether you're on a locked-down system or not.


carolinemodic (approved):
we should probably test this with a secure boot locked down image if you haven't already

config/admin-functions/expand-var-filesystem.sh
```diff
@@ -0,0 +1,47 @@
...
+# Deteremine the volume to extend based on whether encrypted /var is used or not
+if [[ $var_map =~ "var_decrypted" ]]; then
+  volume_to_extend=$( grep var_decrypted /etc/crypttab | awk '{print $2}' )
+else
+  volume_to_extend=$( lvdisplay ${var_map} | grep "LV Path" | awk '{print $3}' )
+fi
+
+# Vx convention: we always expect partition 3
+# if this changes in the future, we'll need to add logic to detect
+growpart "/dev/${parent_partition}" 3 || true
```
benadida, line 33: is this assumption necessary? Can we find the partition number from the label? I can imagine this being a silent failure if we expand the wrong partition and we don't find out until scale testing that `/var` is tiny.

config/admin-functions/expand-var-filesystem.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# LVM_SYSTEM_DIR is necessary for locked down images since the default of /etc/lvm is 
+# read-only in a locked down image
+LVM_SYSTEM_DIR=/home/.lvm pvresize $pvs_path
+
+# This extends the logical volume to the max available space (as created by the previous
+# commands). We pass along the "insecure" passphrase for systems using encrypted /var,
+# and it has affect on systems not using encrypted /var
```
benadida, line 41:
```suggestion
# and it has no effect on systems not using encrypted /var
```

I think you meant that this has *no* effect?

config/admin-functions/show-vx-suite-admin-menu.sh
```diff
@@ -12,6 +12,12 @@ if [[ $(tty) = /dev/tty1 ]] && [[ -f "/home/REKEY_VIA_TPM" ]]; then
   sudo "${VX_FUNCTIONS_ROOT}/rekey-via-tpm.sh"
 fi
 
+# We do this after the rekey operation to save time. It's much faster to encrypt /var
+# before we expand it to the max available storage.
+if [[ $(tty) = /dev/tty1 ]]; then
+  sudo "${VX_FUNCTIONS_ROOT}/expand-var-filesystem.sh"
```
benadida, line 18: I'm confused, we're expanding after rekeying, but in the expansion we're using `insecure` as the passphrase?

amcmanus: If it's a non-secure image, we don't rekey, so it uses `insecure` as the passphrase. If it's a secure image, the `crypttab` config will cause it to use the TPM instead, and the passphrase will be ignored.

benadida: ahhh, right, thank you.

config/admin-functions/expand-var-filesystem.sh
amcmanus: I didn't see an easy/obvious way to do this, but I also didn't spend too much time on it since our preseeds for both vxdev and production result in a partition scheme where `/boot/efi` is on 1, `/boot` is on 2, and everything else is on 3. I can spend some more time on it though to see if there's a good approach that will work across our different system types.

benadida: hmmmm, okay, you make a convincing point that this is likely to be quite stable unless we make a major change. I'm good with it.

PR #395, opened 2024-09-19
Add logic for rebooting to the vendor menu

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/4760

This PR completes the logic for rebooting to the vendor menu, i.e., rebooting to what we currently think of as the TTY2 menu in TTY1. The step-by-step flow is this:

1. Vendor authenticates with vendor card and clicks "Reboot to Vendor Menu"
2. VxSuite backend create a `/vx/config/app-flags/REBOOT_TO_VENDOR_MENU` file as the `vx-services` user and reboots (done in https://github.com/votingworks/vxsuite/pull/5419)
3. VxSuite sees this file flag on boot and automatically logs in as the `vx-admin` user
4. The `vx-admin` user is configured to auto-run the `show-vx-suite-admin-menu.sh` script (existing logic)
5. This script removes the `/vx/config/app-flags/REBOOT_TO_VENDOR_MENU` file so that we don't keep booting to the vendor menu

## Testing

Before merging, will test this manually by editing the code on a non-locked-down production machine

### DIFF
--- a/config/autologin.sh
+++ b/config/autologin.sh
@@ -3,7 +3,7 @@
 # This script is designed to be run in override.conf file for the service getty@tty1.
 # It conditionally selects which user to automatically log in as, based on whether or not the
 # machine 1) needs configuration or 2) is being rebooted into the vendor menu.
-if [[ -f /vx/config/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT ]] || grep -q 1 /vx/config/REBOOT_TO_VENDOR_MENU; then
+if [[ -f /vx/config/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT ]] || [[ -f /vx/config/app-flags/REBOOT_TO_VENDOR_MENU ]]; then
     USER=vx-admin
 else
     USER=vx-ui
--- a/config/autologin.sh
+++ b/config/autologin.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 
 # This script is designed to be run in override.conf file for the service getty@tty1.
-# It conditionally selects which user to autologin based on whether or not the
-# machine needs configuration.
-if [[ -f /vx/config/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT ]]; then
+# It conditionally selects which user to automatically log in as, based on whether or not the
+# machine 1) needs configuration or 2) is being rebooted into the vendor menu.
+if [[ -f /vx/config/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT ]] || grep -q 1 /vx/config/REBOOT_TO_VENDOR_MENU; then
     USER=vx-admin
 else
     USER=vx-ui
--- a/setup-machine.sh
+++ b/setup-machine.sh
@@ -285,6 +285,7 @@ sudo ln -s /vx/code/config/grub.cfg /vx/admin/config/grub.cfg
 
 # machine configuration
 sudo mkdir -p /var/vx/config
+sudo mkdir /var/vx/config/app-flags
 sudo ln -sf /var/vx/config /vx/config
 
 sudo ln -s /vx/code/config/read-vx-machine-config.sh /vx/config/read-vx-machine-config.sh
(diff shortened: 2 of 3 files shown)


setup-machine.sh
```diff
@@ -346,10 +346,13 @@ sudo chown -R vx-services:vx-services /var/vx/data
...
 
-# config readable & executable by all vx users, writable by admin.
+# Config is writable by the vx-admin user and readable/executable by all vx-* users.
+# /vx/config/REBOOT_TO_VENDOR_MENU is special-cased to be writable by all vx-* users.
+sudo echo 0 > /var/vx/config/REBOOT_TO_VENDOR_MENU
 sudo chown -R vx-admin:vx-group /var/vx/config
 sudo chmod -R u=rwX /var/vx/config
-sudo chmod -R g=rX /var/vx/config
+sudo chmod -R g=rX  /var/vx/config
+sudo chmod    g=rwX /var/vx/config/REBOOT_TO_VENDOR_MENU
```
arsalansufi, line 355: One subtlety here is that I’m granting the whole group write permissions on this file, so vx-admin, vx-services, and vx-ui, when really only the first 2 need that access. I figured not a deal breaker but happy to find a way to make the permission more targeted to just those two users, if need be

config/autologin.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 
 # This script is designed to be run in override.conf file for the service getty@tty1.
-# It conditionally selects which user to autologin based on whether or not the
-# machine needs configuration.
-if [[ -f /vx/config/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT ]]; then
+# It conditionally selects which user to automatically log in as, based on whether or not the
+# machine 1) needs configuration or 2) is being rebooted into the vendor menu.
+if [[ -f /vx/config/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT ]] || grep -q 1 /vx/config/REBOOT_TO_VENDOR_MENU; then
```
arsalansufi, line 6: The only remaining follow-up, which I’ll ticket separately and have discussed with Adam, is removing TTY2 access altogether on locked-down images

setup-machine.sh
benadida: There's a defense-in-depth argument here to not let vx-ui do this, but I'm okay with this being a followup ticket.

benadida (approved):
Unsure why this is using a diff pattern than `RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT` which just looked for presence of the file. I don't mind too much, but the difference seems unnecessary.

arsalansufi:
Ah yes @benadida I should've left a comment explaining why the difference is necessary. If I were to just use the existence of the file, I'd have to give `vx-services` greater permissions, specifically write permissions on the `/vx/config` directory, since creating a file in a directory requires write permissions on the directory. So to keep the permissions as granular as possible, I chose to pre-create the file and only give `vx-services` the ability to edit that one file instead of create (any) new files.

benadida:
> Ah yes @benadida I should've left a comment explaining why the difference is necessary. If I were to just use the existence of the file, I'd have to give `vx-services` greater permissions, specifically write permissions on the `/vx/config` directory, since creating a file in a directory requires write permissions on the directory. So to keep the permissions as granular as possible, I chose to pre-create the file and only give `vx-services` the ability to edit that one file instead of create (any) new files.

I don't feel super strongly about this, but one way we could approach this is move all flags to `/vx/config/flags/` and just give permission to that one directory.

But again, I'm okay with this.

arsalansufi:
Not a bad idea! We have some flags that should only be toggled by `vx-admin`, and not `vx-services`. But we could address that by naming the flag directory specifically, e.g., `/vx/config/app-flags`, and only putting certain flags in it. I agree that a file existence check is easier than a file contents check so will make the change

amcmanus:
@arsalansufi, disabling switching to other ttys is a simple change.

arsalansufi:
Ah that's great @amcmanus! That's enough info for me to open my own separate PR!

amcmanus (approved):

PR #396, opened 2024-09-20
Disable terminal access via TTY2 for production images

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/4760

The last step in the transition to vendor cards, this PR disables terminal access via TTY2 for production images. On production images, the only way to get to the former TTY2 admin menu is now by using a vendor card and rebooting into that menu in TTY1.

Thanks to @amcmanus [for the assist](https://github.com/votingworks/vxsuite-complete-system/pull/395#issuecomment-2360982582)!

I initially thought that I might add the TTY2 removal to the lockdown.sh script but thinking some more decided to just incorporate it into setup-machine.sh, making it yet another behavior that's toggled based on your decision to make a QA image vs. a production image.

It's increasingly the case that we're either making QA images or locked-down production images, not in-betweens. A reminder that a terminal can still be accessed on a locked-down production image by using a vendor card and ctrl-c-ing out of the admin menu.

## Testing

Tested the config file manually on a production-imaged ThinkPad


setup-machine.sh
```diff
@@ -54,18 +54,20 @@ MODEL_NAME=${MODEL_NAMES[$CHOICE_INDEX]}
...
-    echo "Using password insecure for vx-admin user."
+    echo "OK, creating a QA image with sudo privileges for the vx-admin user and terminal access via TTY2."
+    echo "Using password insecure for the vx-admin user."
 else
+    IS_QA_IMAGE=0
     VXADMIN_SUDO=0
     echo "Ok, creating a production image. No sudo privileges for anyone!"
     echo
-    echo "Next, we need to set the admin password for this machine."
+    echo "Next, we need to set a password for the vx-admin user."
```
arsalansufi, line 70: I _think_ as of this change we may actually now be in a place where the vx-admin user doesn't need a password

setup-machine.sh
```diff
@@ -54,18 +54,20 @@ MODEL_NAME=${MODEL_NAMES[$CHOICE_INDEX]}
 echo "Excellent, let's set up ${CHOICE}."
 
 echo
-read -p "Is this image for QA where you want sudo privileges, the ability to record screengrabs, etc.? [y/N] " qa_image_flag
+read -p "Is this image for QA where you want sudo privileges, terminal access via TTY2, the ability to record screengrabs, etc.? [y/N] " qa_image_flag
 
 if [[ $qa_image_flag == 'y' || $qa_image_flag == 'Y' ]]; then
+    IS_QA_IMAGE=1
     VXADMIN_SUDO=1
```
benadida, line 61: can we kill this var `VXADMIN_SUDO` now?

benadida (approved):

setup-machine.sh
arsalansufi: Good catch! Will switch over to `IS_QA_IMAGE`

PR #398, opened 2024-10-07
Update set clock script to support all timezones

carolinemodic:
We previously hit an issue due to the fact that this script did not distinguish between Mountain (Phoenix) which is standard time year round and Monutain (Denver) which has daylight savings. Updates the script to include both mountain time options explicitly and add in other US timezones we didn't have for Hawaii/Alaska to future proof. This script now matches the time zone options exposed in the app when setting the time.


arsalansufi (approved):
The nuances of timezones haha, had no idea Phoenix was different!

PR #400, opened 2024-10-10
add reboot to BIOS to vendor menu

adghayes:
## Overview

Adds a reboot to BIOS option to the vendor menu. Tested in my VM. Using the same command we currently use in `vxsuite` to do the reboot.

## Before

Screen Shot 2024-10-10 at 7 55 28 AM

## After

Screen Shot 2024-10-10 at 7 54 33 AM


adghayes:
@arsalansufi I think we should rename the whole `vx-admin` user and config files to use the word `vendor`, but since the script are referenced I didn't want to dig in too much right now and make a mistake that'll hold up the next build, but I certainly can (it's probably safe, just find and replace)

arsalansufi (approved):
LGTM!

arsalansufi:
Discussed with @adghayes in person and agree with the decision to rename `vx-admin` `vx-vendor`! The user is only used to access the basic configuration wizard and the vendor menu, both vendor-specific.

And it'd be nice to better disambiguate VxAdmin the machine from `vx-admin` the system user

PR #403, opened 2024-10-29
rename "admin" entities to "vendor" entities

adghayes:
We've introduced the notion of vendor cards and vendor login in the complete system, but the actual logged in user is still `vx-admin`. This is potentially confusing with our product, "VxAdmin". Because we're only using the `vx-vendor` user for vendor tasks AFAIK, it's safe and clearer to rename it `vx-vendor`.


config/sudoers-for-dev
```diff
@@ -18,31 +18,31 @@ Defaults	secure_path="/vx/code/config/admin-functions:/usr/local/sbin:/usr/local
...
+vx-vendor ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/intermediate-scripts/create-cert
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/intermediate-scripts/create-cert
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/auth/src/intermediate-scripts/sign-message
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/backend/src/intermediate-scripts/*
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/vxsuite/libs/usb-drive/scripts/*
-vx-services ALL=(root:ALL) NOPASSWD: /vx/admin/admin-functions/hash-signature.sh
+vx-services ALL=(root:ALL) NOPASSWD: /vx/vendor/vendor-functions/hash-signature.sh
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/pactl.sh
-vx-ui ALL=(root:ALL) NOPASSWD: /vx/code/config/admin-functions/timedatectl
+vx-ui ALL=(root:ALL) NOPASSWD: /vx/code/config/vendor-functions/timedatectl
```
adghayes, line 45: @arsalansufi we do have the backend and the frontend using a couple of "vendor functions" but I think it's still not necessarily confusing. Or I could back out and rename the user `vx-admin` but rename the directory to `/vx/system` and `/vx/system-functions`. Either way I think it's valuable to get away from `admin`.

amcmanus (approved):

config/sudoers-for-dev
arsalansufi: Yeah thanks for flagging! Not too concerned about this either, as I actually want to change the control flow for these cases, moving the relevant function into vxsuite, with the vendor script calling the vxsuite script.

This is how we program sys admin cards in the vendor menu for example. The vendor menu / vxsuite-complete-system calls into vxsuite. I'm trying to avoid the other way around, vxsuite calling into vxsuite-complete-system, moving forward.

And will address these lingering examples at a future time, post cert submission.

arsalansufi (approved):
LGTM! Thank you for covering this

config/sudoers-for-dev
arsalansufi: Speaking of which, we need to update https://github.com/votingworks/vxsuite/blob/f6ed8ec7c85b83bfff666f81dbfacb6d76df8d7d/libs/auth/src/signed_hash_validation.ts#L31, which I'll take care of

I have a hunch that `/vx/code/config/vendor-functions/timedatectl` is no longer used by `vx-ui`, i.e., kiosk-browser

PR #405, opened 2024-10-31
misc v4 fixes

amcmanus:
This PR resolves self-hosted apt repo cleanup issues, along with supporting the new default of using dev private keys in QA images. It also attempts to reduce the image size, but I think this will require additional process to see it reflected in the final compressed image.

### DIFF
--- a/setup-machine.sh
+++ b/setup-machine.sh
@@ -479,7 +479,7 @@ else
     sudo cp config/sudoers /etc/sudoers
 fi
 
-# NOTE AGAIN: no more sudo commands below this line. Privilegs have been removed.
+# NOTE AGAIN: no more sudo commands below this line. Privileges have been removed.
 
 # remove everything from this bootstrap user's home directory
 cd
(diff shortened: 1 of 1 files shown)


setup-machine.sh
```diff
@@ -462,27 +462,37 @@ sudo passwd -l vx-services
...
+# dev VotingWorks cert by writing it to the expected file path
+if [[ "${IS_QA_IMAGE}" == 1 ]] ; then
+    sudo cp \
+        /vx/code/vxsuite/libs/auth/certs/dev/vx-cert-authority-cert.pem \
+        /vx/code/vxsuite/libs/auth/certs/prod/vx-cert-authority-cert.pem
+fi
+
 # copy in our sudoers file, which removes sudo privileges except for very specific circumstances
 # where needed
+# NOTE: you cannot use sudo commands after this runs
```
arsalansufi, line 475: Ah right of course, thanks for correcting!

setup-machine.sh
```diff
@@ -462,27 +462,37 @@ sudo passwd -l vx-services
...
 cd
 rm -rf *
 rm -rf .*
 
+# see if we can reclaim disk space from cache after home directory deletions
+/usr/bin/sync
+cd /tmp
+ls
+cd
+ls -altr
```
arsalansufi, line 494: Ooh nice excited to see this in action!

PR #409, opened 2024-11-02
Use nice code version strings for official release images

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/vxsuite/issues/5106

Our app code version strings currently look something like this: `2021.03.29-d34db33fcd`. These strings are great from a record-keeping perspective and great for intermediate QA images, but not ideal for final release images, which should use our v3.1.2, v4.0.0, etc., terminology.

For the recent v3.1.* patches, I manually edited the `code-version` file to accomplish this.

This PR adds a prompt to the setup-machine script, asking if an image is an official release image, after you confirm that the image is not a QA image.

If I am to diagram this:

```
Images
├── QA images
└── Production images
    └── Official release images --> Only these will use the nice code version strings
```

For complete clarity, any of the above can be locked-down. Whether an image is locked-down or not has no bearing on the code version string.

I don't want to always use the nice code version strings because I don't want us to ever mistake a QA image or a preliminary production image for an official release image. We should only ever be creating one set of images for a given version string. And it'll likely only ever be SLI who says yes to the `Is this additionally an official release image? [y/N]` prompt, in the context of Trusted Build.

Just to be safe, even when you answer yes to this prompt, I keep the full version string under the hood at `/vx/code/code-version-full`, for us to reference just in case.

## Sample prompt sequences

_QA image_

```
Is this image for QA, where you want sudo privileges, terminal access via TTY2, and the ability to record screengrabs? [y/N] y
OK, creating a QA image with sudo privileges for the vx-vendor user and terminal access via TTY2.
Using password insecure for the vx-vendor user.
```

_Production image, not an official release image_

```
Is this image for QA, where you want sudo privileges, terminal access via TTY2, and the ability to record screengrabs? [y/N] n
Ok, creating a production image. No sudo privileges for anyone!

Is this additionally an official release image? [y/N] n
OK, not an official release image.

Next, we need to set a password for the vx-vendor user.
Set vx-vendor password: 
Confirm vx-vendor password: 
Password confirmed.
```

_Production image, not an official release image, mistaken yes with switch to no at confirm_

```
Is this image for QA, where you want sudo privileges, terminal access via TTY2, and the ability to record screengrabs? [y/N] n
Ok, creating a production image. No sudo privileges for anyone!

Is this additionally an official release image? [y/N] y
Are you sure? [y/N] n
OK, not an official release image.

Next, we need to set a password for the vx-vendor user.
Set vx-vendor password: 
Confirm vx-vendor password: 
Password confirmed.
```

_Official release image_

```
Is this image for QA, where you want sudo privileges, terminal access via TTY2, and the ability to record screengrabs? [y/N] n
Ok, creating a production image. No sudo privileges for anyone!

Is this additionally an official release image? [y/N] y
Are you sure? [y/N] y
OK, will set the displayed code version to: v4.0.0

Next, we need to set a password for the vx-vendor user.
Set vx-vendor password: 
Confirm vx-vendor password: 
Password confirmed.
```


setup-machine.sh
```diff
@@ -65,6 +65,13 @@ else
     IS_QA_IMAGE=0
     echo "Ok, creating a production image. No sudo privileges for anyone!"
     echo
+    read -p "Is this additionally an official release image? [y/N] " release_image_flag
+    if [[ "${release_image_flag}" == 'y' || "${release_image_flag}" == 'Y' ]]; then
+        IS_RELEASE_IMAGE=1
+        VERSION="$(< VERSION)"
```
benadida, line 71: You might want a confirmation in case of typo.

benadida (approved):

setup-machine.sh
arsalansufi: Great call, updated and added some sample prompt sequences to the PR description

arsalansufi:
@mattroe Your pet peeve is addressed! 😝

PR #411, opened 2024-11-07
Fix pulseaudio permissions

amcmanus:
The `~vx-ui/.config/pulse` directory ended up being owned by `root` rather than the `vx-ui` user. This prevents a symlink being created during pulseaudio initialization that is needed for `vx-ui` and `vx-services` users to access the audio config. This PR fixes that permissions issue.


arsalansufi (approved):
Ah yes this makes perfect sense, thanks for fixing so quickly!

PR #415, opened 2024-11-12
Outwardly refer to VxMarkScan as VxMark and remove VxMark(Old) as option

arsalansufi:
Issue link: https://github.com/votingworks/vxsuite/issues/5588

Quite a while back, we decided to rename VxMarkScan VxMark and now think of the v2 VxMark deployed in Mississippi as VxMark(Old). We'll eventually come up with a better naming scheme and more fully rename things in code.

But before we submit v4 for cert, let's at least remove VxMark(Old) as an option in the `setup-machine.sh` script and outwardly refer to VxMarkScan as VxMark.

This will hopefully avoid some confusion during SLI Trusted Build.

### DIFF
--- a/config/vendor-functions/show-vendor-menu.sh
+++ b/config/vendor-functions/show-vendor-menu.sh
@@ -108,9 +108,6 @@ while true; do
   echo "${#CHOICES[@]}. Recreate Machine Cert"
   CHOICES+=('recreate-machine-cert')
 
-  echo "${#CHOICES[@]}. Reset System Authentication Code"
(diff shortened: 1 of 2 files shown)


adghayes (approved):

benadida (approved):

arsalansufi:
Noting that, while I was at is, I also removed a no longer relevant vendor menu option (Reset System Authentication Code): https://github.com/votingworks/vxsuite-complete-system/pull/415/commits/8a30d845dfa9d4a75bdff84612d00c196abebf22

PR #420, opened 2024-11-19
Bump vxsuite for RC1 image creation (!!) (and other final edits)

arsalansufi:
Opening this PR now even though I'm waiting until west-coast EOD for a final change to potentially land, just so I can proactively get it approved before people are less online.

Will pull in any final changes before actually merging and creating images!


benadida (approved):
Remember to update before merge:)

PR #421, opened 2024-11-20
Reduce boot messages and final size of VM

amcmanus:
This PR suppresses or redirects various boot sequence output via easily available options. In the future, we will further streamline the boot sequence via hardware-specific settings and a consistent boot logo experience.

This PR also includes a one-time use of the `fstrim` tool to reclaim disk space within a VM after `setup-machine` initiates a reboot or shutdown of the system. This results in a significant reduction in the size of our final images, making transferring them to/from S3 and USB drives much faster.

### DIFF
--- a/config/autologin.sh
+++ b/config/autologin.sh
@@ -10,4 +10,4 @@ else
 fi
 # For the getty service, you can't have any nested processes, thus we use exec to
 # make mingetty take over the process of this script.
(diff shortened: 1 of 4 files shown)


benadida (approved):

arsalansufi (approved):
LGTM!

PR #425, opened 2024-11-26
suppress non-critical kernel boot messages in a locked down image

amcmanus:
It's a bit of a hammer, but this will remove non-critical boot messages in a locked down image. We should still implement a better splash screen solution that works for all our hardware options, but this improves the overall boot experience.


arsalansufi (approved):
Nice!

arsalansufi:
Also just noting that I've flagged this as a good contender for a set of v4.0.1 mods that we could present to SLI before the NH cert campaign finishes: https://github.com/votingworks/vxsuite/issues/5638

PR #437, opened 2025-04-22
Update the /var expansion process in support of updating vx-iso functionality

amcmanus:
This PR updates how the automated expansion of the `/var` partition is performed, based on updates to vx-iso.

It is no longer tied to the basic configuration wizard flow, but is instead reliant on an explicit flag set by the vx-iso flash image process. An explicit reboot has also been added since that was built into the previous basic config wizard flow.

### DIFF
--- a/config/vendor-functions/show-vendor-menu.sh
+++ b/config/vendor-functions/show-vendor-menu.sh
@@ -12,8 +12,15 @@ if [[ $(tty) = /dev/tty1 ]] && [[ -f "/home/REKEY_VIA_TPM" ]]; then
   sudo "${VX_FUNCTIONS_ROOT}/rekey-via-tpm.sh"
 fi
 
-if [[ $(tty) = /dev/tty1 ]] && [[ -f "${VX_CONFIG_ROOT}/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT" ]]; then
+# Note: EXPAND_VAR will be created as part of a vx-iso install
+# This prevents the var expansion from running in VMs while other
+# config/setup may be taking place. We only want to expand var
+# on systems installed to physical hardware.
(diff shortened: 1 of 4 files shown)


arsalansufi (approved):

config/vendor-functions/show-vendor-menu.sh
```diff
@@ -12,8 +12,21 @@ if [[ $(tty) = /dev/tty1 ]] && [[ -f "/home/REKEY_VIA_TPM" ]]; then
   sudo "${VX_FUNCTIONS_ROOT}/rekey-via-tpm.sh"
 fi
 
-if [[ $(tty) = /dev/tty1 ]] && [[ -f "${VX_CONFIG_ROOT}/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT" ]]; then
+# Note: EXPAND_VAR will be created as part of a vx-iso install
+# This prevents the var expansion from running in VMs while other
+# config/setup may be taking place. We only want to expand var
+# on systems installed to physical hardware.
+if [[ $(tty) = /dev/tty1 ]] && [[ -f "${VX_CONFIG_ROOT}/EXPAND_VAR" ]]; then
```
arsalansufi, line 19: Noting that we're going to add an `|| [[ -f "${VX_CONFIG_ROOT}/RUN_BASIC_CONFIGURATION_ON_NEXT_BOOT" ]]` to make sure that, when using a new image with an old vx-iso, we still properly expand the var partition

amcmanus: There was no good way to solve this via the vendor menu / autologin approach, so I instead set the `EXPAND_VAR` flag file via setup-machine.sh. New vx-iso releases also set it during imaging, so we can eventually remove it, but this will make it simple for old vx-iso releases to still work with the new approach.

PR #440, opened 2025-05-14
Update hosts information for vxpollbook machines, kiosk-browser bug fix

amcmanus:
This PR sets hostname information for vxpollbook machines as part of the machine id step in basic configuration.

It also removes the `--add-file-perm` option from kiosk-browser since that option was removed in a recent update.

### DIFF
--- a/config/sudoers
+++ b/config/sudoers
@@ -21,6 +21,7 @@ root	ALL=(ALL:ALL) ALL
 
 # fine-grained sudo permissions for certain users & actions
 vx-vendor ALL=(root:ALL) NOPASSWD: /vx/vendor/vendor-functions/set-clock.sh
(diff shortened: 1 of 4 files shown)


carolinemodic (approved):
Some of the complete-system files like sudoers-for-dev are duplicated right now in build-system for the pollbook use case. I'm not entirely sure the best way to manage those files long term but just a heads up in case you haven't noticed that. Maybe we should move those into complete-system and label them for pollbook or something I'm not sure.

amcmanus:
> Some of the complete-system files like sudoers-for-dev are duplicated right now in build-system for the pollbook use case. I'm not entirely sure the best way to manage those files long term but just a heads up in case you haven't noticed that. Maybe we should move those into complete-system and label them for pollbook or something I'm not sure.

Yeah, I'm leaning toward eventually consolidating as much as possible, but sticking with the current structure for efficiency right now.

PR #447, opened 2025-05-29
add support for changing ELO screen brightness via ddcutil

amcmanus:
This PR adds the configuration necessary to automatically set ELO screen brightness at boot. 

- the `xinitrc` config runs the actual `ddcutil` command as part of initializing our X session
- `ddcutil` requires the `i2c-dev` module be loaded, so that config has been added to `/etc/modules-load.d`, which results in the module automatically loading during boot


arsalansufi (approved):
LGTM! Assuming 70% is the brightness we want - I can't recall where we recorded/documented the desired brightness

arsalansufi:
Ah actually just stumbled across https://votingworks.slack.com/archives/CFSJEQ26N/p1748529767542459

PR #448, opened 2025-06-04
Add paplay.sh app script for playing audio through specific outputs

kofi-q:
Task: https://github.com/votingworks/vxsuite/issues/6484

We're enabling screen reader audio in VxScan, which will play through headphones plugged into an always-connected USB audio device (integrated into a tactile controller).

To enable still being able to play scan success/fail sounds through the speaker, we'll be using the existing `pactl.sh` script to list audio sinks at startup and then using this new `paplay.sh` script to play sounds through the sink corresponding to the speaker from the backend service (e.g. `sudo /vx/code/app-scripts/paplay.sh sucess.mp3 --device=alsa_ouptut.pci...`).

Following the same pattern as `pactl.sh`, assuming this also needs to run as `vx-ui`, but not sure if this covers everything needed for this to work on VxScan.


amcmanus (approved):
As long as the `paplay` functionality is being used in vxsuite the same way as `pactl`, this pattern should work.

PR #449, opened 2025-06-06
Bump both submodules to latest main

arsalansufi:
Prepping for v4.0.2 QA image creation! I'll still wait a bit longer for any other lingering changes to vxsuite before I merge this and make images from it


kofi-q (approved):
🚀

PR #451, opened 2025-06-09
VxPollbook: add support for TPM-backed certificate auth via strongswan

amcmanus:
This PR adds the necessary support for certificate-based authentication (backed by the TPM) between mesh-connected pollbook machines. Of note:

- New keys are created in the TPM. These keys are different than our existing key, and you can learn more in the [Strongswan TPM docs](https://docs.strongswan.org/docs/latest/tpm/tpm2.html)
- The existing cert creation/verification processes were updated to include the use of the new TPM handle. That also required an update in vxsuite, [seen here](https://github.com/votingworks/vxsuite/commit/eaa341ca1cf5bf8b7d0f5d17f76ff5b07553808c).

### DIFF
--- a/config/vendor-functions/create-machine-cert.sh
+++ b/config/vendor-functions/create-machine-cert.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 
 # Requires sudo
+# TODO: With the introduction of pollbook certs, we are duplicating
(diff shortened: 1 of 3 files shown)


arsalansufi (approved):
Some naming suggestions but otherwise LGTM!

PR #453, opened 2025-06-24
Updates our rsyslog template to capture two event types

amcmanus:
Fix for https://github.com/votingworks/cert-discrepancies/issues/156

The log messages needed to be updated to catch these two event types.


arsalansufi (approved):
LGTM! I take it the underlying log messages changed as part of an OS update?

PR #461, opened 2025-07-02
Improve dm-verity boot logging

amcmanus:
In some scenarios, it's possible to not see dm-verity boot related messages shortly after a system is booted, including when logs are exported. This PR changes our default buffer from 100 log lines being processed down to 10 so that logs are written to disk more frequently. In addition, the dm-verity filter has been updated to only include the successful boot state rather than all verity related logs.


arsalansufi (approved):
🪵

PR #462, opened 2025-07-11
Remap tactile controller multimedia keys to corresponding VxSuite keys

kofi-q:
To support the new `AudioNav 1409-34011` tactile controller in the upcoming BMD, we'll need to remap a few button to keyboard events that are used in app to control screen reader volume and speech rate to keep things consistent across controllers. By default, the device emits multimedia key events for those buttons, which trigger OS functions (on VMs), preempt the in-app event handling.

Using `xmodmap`, which appears to already be installed in our base images, to set the mappings for `vx-ui` at startup. Tested with a recent VxScan image, with manual edits to mimic this setup.

~Also adding `app-scripts/xmodmap.sh` as backup, to be run in the server whenever we detect a device connection for the tactile controller. That should handle any potential disconnects from electrical testing.~


config/xinitrc
```diff
@@ -2,6 +2,7 @@ xset -dpms
 xset s off
 xset s noblank
 xrdb ~/.Xresources
+xmodmap ~/.Xmodmap
```
kofi-q, line 5: Just double-checking - is it safe to assume `xmodmap` is available on locked-down prod images?

From searching, it looks like it's bundled with `x11-xserver-utils`, which I don't think we install directly, but `xmodmap` was available on the QA image I tested with, so I'm guessing it's a transitive dependency.

amcmanus (approved):

PR #464, opened 2025-07-29
Run fipsinstall before operations that require openssl

amcmanus:
We (currently) persist the fipsmodule.cnf across installs, but if there are other kernel/system level upgrades between different versions, it can cause the fipsmodule.cnf to be invalid until it's installed/verified again. This PR runs the necessary fips install/verification process before other operations so that we guarantee an updated and valid fips configuration for the newly installed version of an application.


arsalansufi (approved):
The simplest fix 👏

PR #510, opened 2026-04-14
Remove mark-scan specific profile rotation

amcmanus:
During early mark-scan testing/development, we added an automatic screen rotation for terminal sessions. This was purely for convenience and has no bearing on production use. That said, it introduced a potential attack via privileged use of the `tee` command. This PR removes that configuration in favor of security over convenience. If this hardware configuration and screen rotation warrants future support, we can re-enable without the use of a privileged `tee` command. Until then, it's best to simply remove.

### DIFF
--- a/config/mark_scan_admin_bash_profile
+++ b/config/mark_scan_admin_bash_profile
@@ -1,6 +1,0 @@
-export PATH=/vx/code/config/vendor-functions:${PATH}
-
-while true; do
-    bash /vx/vendor/vendor-functions/show-vendor-menu.sh
-done
-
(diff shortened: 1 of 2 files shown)


benadida (approved):

arsalansufi:
I think we also more importantly need to remove this: https://github.com/votingworks/vxsuite-complete-system/blob/9a6533150933eac10a02108ca012e25b9c22184e/config/sudoers#L42

amcmanus:
> I think we also more importantly need to remove this:
> 
> https://github.com/votingworks/vxsuite-complete-system/blob/9a6533150933eac10a02108ca012e25b9c22184e/config/sudoers#L42

We do, but there are some other references to the `tee` command, so I want to be sure before removing here.

PR #511, opened 2026-04-14
sanitize vendor menu input to prevent bash injection vulnerability

amcmanus:
This PR addresses a potential attack vector identified via the vendor menu. With a specifically crafted menu input, an attacker could gain console access as the vx-vendor user and their associated privileges. While that is protected by additional limitations, we still want to prevent this entry point. By sanitizing the input, we can prevent the console access. 

I went with a regex check instead of the recommended `printf` approach since it was accepting octal, hex, and other inputs (including leading whitespaces). I did not create actual attacks via that approach, but it seemed better to eliminate it entirely.

### DIFF
--- a/config/vendor-functions/show-vendor-menu.sh
+++ b/config/vendor-functions/show-vendor-menu.sh
@@ -127,7 +127,13 @@ while true; do
   echo
   read -p "Select menu item: " CHOICE_INDEX
 
-  CHOICE=${CHOICES[$CHOICE_INDEX]}
+  # Protect from malicious input, only accept 0-99 as input
+  if ! [[ "$CHOICE_INDEX" =~ ^[0-9]{1,2}$ ]]; then
+    CHOICE="invalid"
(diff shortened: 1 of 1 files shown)


benadida (approved):
looks good. Do we also need a bounds check?

amcmanus:
> looks good. Do we also need a bounds check?

What kind of bounds checking are you thinking?

benadida:
> What kind of bounds checking are you thinking?

on the `CHOICES` array so you can't enter a number larger than the array?

amcmanus:
> > What kind of bounds checking are you thinking?
> 
> on the `CHOICES` array so you can't enter a number larger than the array?

I think that's already covered by the new regex and checking available options in the array. If you enter a 0-99 number that doesn't map to an option in the array, you fall through and the menu resets. We can make it very explicit rather than relying on the catch-all if you think that's preferable.

PR #513, opened 2026-04-16
add console session option to qa images

amcmanus:
Adds an option to the vendor menu to open a console session from the vendor menu in QA images to assist with troubleshooting.

### DIFF
--- a/config/vendor-functions/show-vendor-menu.sh
+++ b/config/vendor-functions/show-vendor-menu.sh
@@ -121,6 +121,9 @@ while true; do
 
     echo "${#CHOICES[@]}. Copy all Screen Recordings to USB"
     CHOICES+=('copy-recordings')
+
+    echo "${#CHOICES[@]}. Console Session"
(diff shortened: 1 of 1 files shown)


benadida (approved):

config/vendor-functions/show-vendor-menu.sh
```diff
@@ -246,6 +249,10 @@ while true; do
       sudo "${VX_FUNCTIONS_ROOT}/copy-recordings.sh"
       read -s -n 1
     ;;
+
+    console-session)
+      sh -c "/bin/bash"
```
benadida: is there any additional check we can have here to check that we are on a QA image before running `/bin/bash` ?

amcmanus: Good question. We can check the qa flag file independently of how it's checked via `read-vx-machine-config.sh` again here. That way, even if this option were somehow executed from a non-qa image, it'd have to pass the check again.

PR #515, opened 2026-05-05
Add local networking to support multi-station adjudication

amcmanus:
This PR adds system level configuration in support of https://github.com/votingworks/vxsuite-build-system/pull/232.


carolinemodic (approved):
We should probably make vxsuite be latest main not my branch, I can make another PR to bump main again once I merge the changes on that branch

PR #517, opened 2026-05-14
Adam/remove vendor functions symlink references

amcmanus:
This PR removes the use of symlinks when calling vendor function scripts.


arsalansufi (approved):
LGTM!

PR #519, opened 2026-05-14
add multi user support to audio scripts

amcmanus:
In production, audio is played via the `vx-ui` user. In dev environments, the local user may differ, and the hardcoded path can cause the application to not play audio or crash. This checks for the expected `vx-ui` user, falls back to the `vx` user, and then ultimately falls back to the current user calling the script.


kofi-q (approved):
🎉  LGTM!

amcmanus:
This will close https://github.com/votingworks/vxsuite/issues/7304

PR #522, opened 2026-05-18
Remove no longer relevant entries from sudoers

arsalansufi:
We have a few entries in the sudoers file that are no longer relevant. They also all correspond to base binaries rather than wrapper scripts. We generally avoid the former in the sudoers file to prevent usages with unrestricted args.

Best to remove these entries for security hygiene 🧹🔒


config/sudoers
```diff
@@ -55,6 +53,4 @@ vx-services ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/paplay.sh
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/set-usb-port-status.sh
 vx-ui ALL=(root:ALL) NOPASSWD: /vx/code/config/vendor-functions/timedatectl
 vx-ui ALL=(root:ALL) NOPASSWD: /usr/bin/brightnessctl
-vx-ui ALL=(root:ALL) NOPASSWD: /usr/local/bin/tpm2-totp
-vx-ui ALL=(root:ALL) NOPASSWD: /bin/efibootmgr
```
benadida, line 59: we don't need this one to reboot to BIOS?

benadida (approved):
one question, otherwise good

config/sudoers
arsalansufi: Nope! The vx-services user now handles that, via this script: https://github.com/votingworks/vxsuite/blob/main/libs/backend/intermediate-scripts/reboot-to-bios

PR #523, opened 2026-05-18
Use wrapper script to restrict logrotate args

arsalansufi:
A little more sudoers cleanup. Here, I'm removing the sudoers entry that grants access to `/sbin/logrotate` with unrestricted args and replacing it with an entry that grants access to a script that performs our exact log rotation command.

Committed `config/vendor-functions/rotate-logs.sh` with executable permissions. `setup-machine.sh` may already handle setting executable permissions, but all other scripts in `config/vendor-functions` also already have this bit set so I'm matching for consistency.

I'll also acknowledge that `config/vendor-functions` isn't the right directory name here as this is a vx-ui action, not a vx-vendor action. But this is our designated location for these scripts, and we already have other scripts in there specific to vx-ui. We can rename later.

### DIFF
--- a/config/sudoers
+++ b/config/sudoers
@@ -53,4 +53,4 @@ vx-services ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/paplay.sh
 vx-services ALL=(root:ALL) NOPASSWD: /vx/code/app-scripts/set-usb-port-status.sh
 vx-ui ALL=(root:ALL) NOPASSWD: /vx/code/config/vendor-functions/timedatectl
 vx-ui ALL=(root:ALL) NOPASSWD: /usr/bin/brightnessctl
-vx-ui ALL=(root:ALL) NOPASSWD: /sbin/logrotate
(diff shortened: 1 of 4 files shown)


benadida (approved):
looks right to me but would love @amcmanus 's confirmation of the options on the shell script.

arsalansufi:
Yeah agreed want @amcmanus's eyes on that bit! I generally assumed that `set -euo pipefail` as well as the traps are safe as the current script that calls `/sbin/logrotate --force /etc/vx-logs.logrotate` also sets those:

https://github.com/votingworks/vxsuite-complete-system/blob/aa1726fc4ef394963e7c91e90cc6049b07e12201/run-scripts/run-kiosk-browser-forever-and-log.sh#L3-L5

arsalansufi:
I'm gonna merge to get closer to ready for builds but can obviously make edits as Adam reviews tomorrow

amcmanus:
Looks right to me.

PR #529, opened 2026-06-03
[DRAFT] v4.0.6 ctrl-c protection test

arsalansufi:
Using this branch (based on the v4.0.6-release-branch) to create some test images, and using this draft PR for discussion

### DIFF
--- a/config/admin_bash_profile
+++ b/config/admin_bash_profile
@@ -1,8 +1,9 @@
-trap '' SIGINT SIGTSTP SIGQUIT
+trap 'logout' SIGINT SIGTSTP SIGQUIT EXIT
 
 export PATH=/vx/code/config/vendor-functions:${PATH}
 
 while true; do
-    bash /vx/code/config/vendor-functions/show-vendor-menu.sh
+    exec /vx/code/config/vendor-functions/show-vendor-menu.sh
 done
 
+logout
--- a/config/admin_bash_profile
+++ b/config/admin_bash_profile
@@ -3,8 +3,6 @@ stty -isig
 
 export PATH=/vx/code/config/vendor-functions:${PATH}
 
-while true; do
-    exec /vx/code/config/vendor-functions/show-vendor-menu.sh
-done
+exec /vx/code/config/vendor-functions/show-vendor-menu.sh
 
 logout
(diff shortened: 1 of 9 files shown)


config/admin_bash_profile
```diff
@@ -1,6 +1,10 @@
+trap 'logout' SIGINT SIGTSTP SIGQUIT EXIT
+stty -isig
```
arsalansufi, line 2: Do we still need these bash profiles now that we're not using base bash as the startup shell?

config/admin_bash_profile
```diff
@@ -1,6 +1,10 @@
...
+
 export PATH=/vx/code/config/vendor-functions:${PATH}
 
 while true; do
```
arsalansufi: Assuming there is still a reason we need these bash profiles, I'd say we should keep the bash profiles and startup shells consistent and remove the `while true; do` here too

config/vx-ui-shell
```diff
@@ -0,0 +1,12 @@
...
+trap 'logout' SIGINT SIGTSTP SIGQUIT EXIT
+stty -isig
+
+export PATH=/vx/code/config/vendor-functions:${PATH}
+
+if [[ $(tty) = /dev/tty1 ]]; then
+   exec startx > /tmp/vx-ui-startx.log 2>&1
+fi
+
+logout
```
arsalansufi, line 12: Can you remind me: What is the purpose of this `logout`?

arsalansufi (commented):
A few questions for discussion while we await the test build

config/admin_bash_profile
amcmanus: I think we should keep them in place, simply as a defense in depth measure.

config/admin_bash_profile
amcmanus: Agreed, will update.

config/vx-ui-shell
amcmanus: Just more defense in depth. If we somehow got past that `startx` command, let's end the session via `logout`

arsalansufi:
Testing looking good and now shifting back to https://github.com/votingworks/vxsuite-complete-system/pull/528!

PR #537, opened 2026-07-10
Use sync instead of sync $MOUNTPOINT

arsalansufi:
A quick follow-up to https://github.com/votingworks/vxsuite-complete-system/pull/536. I had a slight misunderstanding when opening that PR. There are three permutations of sync commands.
1. sync
2. sync -f PATH
3. sync PATH

@amcmanus and I were discussing how `sync -f PATH` may behave in unspecified ways when we have symlinks and multiple mount points for a USB drive. But `sync` will flush across all locations. I mistakenly switched to the intermediate `sync PATH` which appears to be the most narrowly scoped option. It only flushes the one item at `PATH` and nothing else. If that's a directory, that may not even include the files in the directory.

My testing has been quite haphazard here.  In one test, I found that data was not synced after `sync -f PATH` but was after `sync PATH`. But that my just have been a fluke. The first test was on a prod image and the second on a QA image. Ultimately, the only option that feels truly safe in this symlink, multiple-mount-point context is `sync`. I also added a sleep for good measure since this code is so one-off and speed is not a priority.

### DIFF
--- a/app-scripts/unmount-usb.sh
+++ b/app-scripts/unmount-usb.sh
@@ -12,5 +12,17 @@ if ! [[ $# -eq 0 ]]; then
 fi
 
 MOUNTPOINT=/media/vx/usb-drive
-sync $MOUNTPOINT
+
+# Pull out all the stops to make sure that data is flushed.
+#
+# The app is now auto-mounting USB drives at device-specific paths, e.g.,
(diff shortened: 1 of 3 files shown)


arsalansufi:
Merging but feel free to leave thoughts retroactively @amcmanus!

PR #540, opened 2026-07-23
disable new pnpm v9 version management at runtime

amcmanus:
pnpm v9+ has a new default for `packageManager` versioning that tries to synchronize versions to the global version. Since we run the pnpm build step offline, only the versions available during the initial `pnpm install --frozen-lockfile` step are available. While the overall build succeeds, running the application in an offline environment will fail because pnpm attempts to resolve version discrepancies via a network connection. Setting this configuration option bypasses this new behavior, while ensuring the application runs with the locked/frozen configuration we expect.

### DIFF
--- a/setup-machine.sh
+++ b/setup-machine.sh
@@ -477,6 +477,17 @@ if [[ "${IS_QA_IMAGE}" == 1 ]] ; then
         /vx/code/vxsuite/libs/auth/certs/prod/vx-cert-authority-cert.pem
 fi
 
+# pnpm v9 and 10 default to managing the package manager version at runtime
+# If there is a difference in what is defined in package.json, it attempts 
+# to add the global version. Since we have no network during the offline phase,
+# it's possible to have libraries / apps with a different version. 
+# The install and build are still succcessful, but the application will not 
+# start because of this attempt to add an already existing pnpm version.
+# This option bypasses the new default behavior since we already use locked down
+# versions of pnpm and do not have any external network connectivity.
+# NOTE: This behavior changes in v11 and will require another update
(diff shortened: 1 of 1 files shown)


setup-machine.sh
```diff
@@ -477,6 +477,17 @@ if [[ "${IS_QA_IMAGE}" == 1 ]] ; then
         /vx/code/vxsuite/libs/auth/certs/prod/vx-cert-authority-cert.pem
 fi
 
+# pnpm v9 and 10 default to managing the package manager version at runtime
+# If there is a difference in what is defined in package.json, it attempts 
+# to add the global version. Since we have no network during the offline phase,
```
arsalansufi, line 482: > If there is a difference in what is defined in package.json, it attempts to add the global version.

Difference between the defined pnpm version in the package.json and what other source exactly? Worth clarifying in the comment

arsalansufi (approved):
LGTM!

arsalansufi:
Also flagging for myself that I'll want to incorporate this into the v4.1.0 tag and share it with SLI in a modified source code drop before the v4.1.0 Trusted Build

setup-machine.sh
```diff
@@ -477,6 +477,17 @@ if [[ "${IS_QA_IMAGE}" == 1 ]] ; then
...
+# it's possible to have libraries / apps with a different version. 
+# The install and build are still succcessful, but the application will not 
+# start because of this attempt to add an already existing pnpm version.
+# This option bypasses the new default behavior since we already use locked down
+# versions of pnpm and do not have any external network connectivity.
+# NOTE: This behavior changes in v11 and will require another update
```
eventualbuddha, line 488: Linking to the pnpm v11 setting for when we come back to update this: https://pnpm.io/settings/cli#pmonfail

eventualbuddha (approved):

PR #545, opened 2026-08-19
Fix kiosk-browser reboot

### LINKED ISSUE #8876: Auto-restart on crash not working
The auto-reboot after 10 mins on the something went wrong screen seemed not to work, at least in this one case. Maybe kiosk-browser had crashed

Issue automatically created from Slack feedback reported by arsalan: 
https://votingworks.slack.com/archives/CJU9MSC6S/p1783473224327449

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/8876

We use the kiosk-browser reboot method to auto-restart the app when it completely crashes and we can't use the backend. This functionality has been broken since https://github.com/votingworks/vxsuite-complete-system/pull/528. As we shifted from a generic bash shell on login to a custom shelll, /usr/sbin and /sbin (which house reboot) were dropped from the vx-ui user's PATH. So calling reboot resulted in a command-not-found error.

When we were just using a generic bash shell on login, while it looks like we were adding /sbin to the path via the vx-ui bash profile

https://github.com/votingworks/vxsuite-complete-system/blob/01ffdf7227dd5ffd84b1bc57df10e8db5f74097f/config/ui_bash_profile#L10

that line was never running because the exec above expectedly wouldn't terminate. /usr/sbin and /sbin were likely getting added to the path via other default on-bash-startup mechanisms.

Just to avoid confusion, I decided to fully delete the vx-vendor and vx-ui bash profiles since the custom startup shells now render those unused. There were a few lingering uses I had to take care of in VxDev and [VxPollBook](https://github.com/votingworks/vxsuite-build-system/pull/268).

## Testing

Will be performing some new builds soon so will test this fully via those


amcmanus (approved):
I'd re-order the `PATH` variable to include `${PATH}` at the end, with `/usr/sbin:/sbin` ahead of it. It's a fairly standard convention, and it puts system level tools ahead of anything that might show up at the user-specific `PATH` level. Other than that (and even without that change), LGTM.

arsalansufi:
Makes sense! Adjusted order

PR #546, opened 2026-08-21
Copy/symlink Playwright Chromium to expected path

arsalansufi:
## Overview

After merging https://github.com/votingworks/vxsuite/pull/9191, we're seeing the following error when trying to print on a QA image:

> Executable doesn't exist at /var/vx/services/.cache/ms-playwright/chromium_headless_shell-1193/chrome-linux/headless_shell

The issue here is that Playwright is installing Chromium in the vx user's home directory whereas, at runtime, we're looking for it in the vx-services user's home directory. The vx user and corresponding home directory are also completely removed by the end of image creation. To remedy this, I'm copying and symlinking Chromium to the expected path.

## Testing

- [x] Tested via an image

### DIFF
--- a/run-scripts/run-admin.sh
+++ b/run-scripts/run-admin.sh
@@ -16,6 +16,7 @@ if [ -z "${ADMIN_WORKSPACE:-}" ]; then
   exit 1
 fi
 
-export PIPENV_VENV_IN_PROJECT=1
 export NODE_ENV=production
+export PIPENV_VENV_IN_PROJECT=1
+export PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS=1
 (trap 'kill 0' SIGINT SIGHUP; make -C vxsuite/apps/admin/backend run & make -C vxsuite/services/converter-ms-sems run & make -C vxsuite/apps/admin/frontend run) | logger -S 4096 --tag votingworksapp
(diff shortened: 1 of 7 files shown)


amcmanus (approved):
Is there a reason to copy and symlink rather than simply copy it to `/vx/services/.cache/ms-playwright` directly?

arsalansufi:
Deleted my previous response because I got my wires crossed for sec. Let me take a sec to refresh my memory on my thinking re read/write perms and symlink vs. direct copy as a result

arsalansufi:
Okay, so my thinking was that the Chromium install directory need not be writable by vx-services at runtime, so better from a security perspective to write it to read-only /vx/code and symlink it to the vx-services home directory.

That said, it's hard to know whether it needs to be writable or not (e.g. what if Playwright writes an update to the directory once in a blue moon - would be hard to catch that), and it might be safer to start with it writable by vx-services. Let me do some more research.

arsalansufi:
Hm so I'm going back and forth a bit as copying directly to /vx/services would mean that the Chromium binary isn't included in the dm-verity hash, when I think it should be. I'm now more inclined to leave things as is and do the research to ensure that things will work in a locked-down environment.

arsalansufi:
Did some investigation and with `PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS=1` we should be safe, but I'll do some additional testing next time we make images

PR #548, opened 2026-08-27
Ensure that VxMark HWTA headphone volume setting is respected

### LINKED ISSUE #9085: VxMark HWTA: Headphone audio is starting too loud despite initialization code explicitly lowering volume
Here's where we explicitly lower volume: https://github.com/votingworks/vxsuite/blob/4fc62e07c007e16162bed6b31af788090f866bc7/apps/mark/backend/src/index.ts#L77-L82

But on boot, I'm still getting the extreme headphone volume.

Interestingly, opening and closing the headphone calibration modal does bring the volume back down to its rightful value. So this volume setting call is working despite the initial one not: https://github.com/votingworks/vxsuite/blob/4fc62e07c007e16162bed6b31af788090f866bc7/apps/mark/frontend/src/electrical_testing/app_root.tsx#L129-L130

arsalansufi:
## Overview

Closes https://github.com/votingworks/vxsuite/issues/9085

The VxMark HWTA headphone volume is currently extremely loud despite the HWTA backend setting the headphone volume to 40%. This appears to be because it's being overridden by a 100% volume set in the xinitrc, which runs with a sleep so happens after the app-level set and takes precedence. The xinitrc 100% volume set isn't needed for the main app either. On VxMark, the app server owns volume setting, setting to 100% for the main app and 40% for the HWTA. The main app uses gain to ensure audio isn't blaringly loud.

## Testing

Tested by patching a QA image and confirming that this does the trick

### DIFF
--- a/config/xinitrc
+++ b/config/xinitrc
@@ -59,10 +59,12 @@ pactl unload-module module-suspend-on-idle
 # (because of course they do)
 #
 # NOTE: On VxMark, the startup chime is played through the speakers from within
-# the app server, so we instead play a shortned version here through the
+# the app server, so we instead play a shortened version here through the
 # headphones, to activate the headphone audio device.
 if cat $VX_CONFIG_ROOT/machine-type | grep "^mark$" > /dev/null; then
-  sleep 2 && aplay ~/chime-short.wav && pactl set-sink-volume 0 100%
+  # On VxMark, the app server owns volume setting so don't set anything here
+  # that might conflict
+  sleep 2 && aplay ~/chime-short.wav
 else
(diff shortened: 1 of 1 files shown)


kofi-q (approved):
🎉  LGTM - something that's worth verifying, that might just be a non-issue, is whether the volume level is enough to activate the headphones (if booting up after it's already set to a lower volume) - remember running into some issues with that while first setting up VxMark. Guessing it's fine in the mainline cases, but just a heads-up

arsalansufi:
Oh good to know! All seems to be working fine in that I'm booting the HWTA (which sets to 40%) and then rebooting. And on reboot, headphones are still working fine, suggesting that the volume level is still sufficient to activate the headphones.

